PR #4, opened 2018-07-19
test: Force running ws container

martinpitt:
Occasionally the VM install fails like this:

```
+ atomic run cockpit/ws
The image 'ws' appears to have not been installed and has an INSTALL label.  You should install this image first.  Re-run with --ignore to bypass this error.
Command '/var/tmp/vm.install' returned non-zero exit status 1.
```

We do run `atomic install`, this seems to be a race condition in Fedora
Atomic. Skip that check, we are not testing `atomic` itself here.


martinpitt:
Example: https://fedorapeople.org/groups/cockpit/logs/pull-3-20180719-082121-1453fb54-cockpit-fedora-atomic/log

PR #6, opened 2018-08-02
Simplify spec file

martinpitt:
- Entirely drop `%build` section, as there is nothing to do anyway.
- Use `%make_install` macro.
- Replace the complicated file list wrangling with a simple directory
  enumeration.

Thanks to Igor Gnatenko for the suggestions!

See https://bugzilla.redhat.com/show_bug.cgi?id=1603146


martinpitt:
This is just a merge from starter-kit. Posting it as a PR to get tests running, but I'll be so daring to self-approve.

PR #8, opened 2018-08-27
Drop python3 build requirement

martinpitt:
Hardcode the package name instead of parsing it from package.json. It
won't change for cockpit-ostree, this was a leftover from starter-kit.


martinpitt:
I already applied this patch downstream in RHEL for cockpit-ostree srpm. CentOS/RHEL 7 don't have Python 3.

martinpitt:
Correct, we don't run the integration tests during RPM build, this is only for upstream and there the `BuildRequires` don't matter.

PR #9, opened 2018-08-27
cockpituous: Automatic release into Fedora

martinpitt:


martinpitt:
Note that the tests don't cover this, but it's still nice to have a current test run right before a release. :-)

PR #16, opened 2019-07-09
test: Set up atomic /var/lib path

marusak:
This is workaround for https://github.com/projectatomic/atomic/issues/1258

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -2,6 +2,9 @@
 # image-customize script to install cockpit/ws container and the locally built RPM
 set -eux
 
+# HACK: https://github.com/projectatomic/atomic/issues/1258
+mkdir -p /var/lib/atomic
+
 # Install cockpit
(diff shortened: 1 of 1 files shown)


test/vm.install
```diff
@@ -2,6 +2,9 @@
 # image-customize script to install cockpit/ws container and the locally built RPM
 set -eux
 
+# HACK: https://github.com/projectatomic/atomic/issues/1258
+export ATOMIC_VAR_LIB=/var/lib/containers/atomic
```
martinpitt: Let's guard this with `if [ "$1" = fedora-atomic ]`, as {rhel,continuous}-atomic don't have that bug and get broken by this.

marusak: this script is called without any arguments so `$1` is not defined. `mkdir -p /var/lib/atomic` had to do.

PR #19, opened 2019-07-11
Drop source map

martinpitt:
They are large and just for debugging. If they are ever needed, they
should move into a separate -debugsource package, but this seems a bit
overkill here.

This reduces the rpm from 680 to 180 kB.


martinpitt:
FTR, after this goes in, I'll make a new cockpit-ostree release. It's overdue.

PR #20, opened 2019-07-11
Add packit configuration

martinpitt:
See <https://packit.dev>. This will gradually replace our cockpituous
configuration.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -1,0 +1,8 @@
+specfile_path: cockpit-ostree.spec
+synced_files:
+  - cockpit-ostree.spec
+upstream_project_name: cockpit-ostree
+downstream_package_name: cockpit-ostree
+actions:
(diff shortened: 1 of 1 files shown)


martinpitt:
This is still a bit hard to test in "dry run" mode. I'm planning to do the next cockpit-ostree release manually with packit (I already disabled the release part ("branch and tag creation") in our webhook), and if that works, switch over the cockpituous config to use packit.

martinpitt:
So it's done. This config file worked by and large. There was some trouble due to shoehorning it into the process without being in the released branch yet, but that hopefully goes away once this PR actually lands.

However, release isn't fully automatic yet. I reported https://github.com/packit-service/packit/issues/410 and https://github.com/packit-service/packit/issues/411 which are now the new blockers for replacing cockpituous.

PR #22, opened 2019-07-22
spec: Depend on package instead of file path

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1731687


martinpitt:
Package name is [different on rhel-atomic](https://209.132.184.41:8493/logs/pull-22-20190722-104545-dbcf25fe-cockpit-project-cockpit-ostree-cockpit-rhel-atomic/log), I pushed a fix.

martinpitt:
Argh, we currently build the rpm in the container, not the VM.

PR #24, opened 2019-07-25
timeline: Use PF4 inspired background color

garrett:


garrett:
This is a super-simple, tiny fix for the background if the icons in the timline. They're set to white; I changed them to use the page's background color.

PR #36, opened 2020-01-30
Modernize build system

martinpitt:


martinpitt:
> Are bots doing these refreshes in this repo?

No, that's what I'm going to enable next.

PR #38, opened 2020-01-30
package.json: Update angular-gettext package dependency

martinpitt:


martinpitt:
Please manually check features related to these files:

```
package.json:    "angular-gettext": "2.4.1",
po/html2po: * Supports the following angular-gettext compatible forms:
po/po.empty.js:        } catch(ex) { console.log(ex); /* Either no angular or angular-gettext */ };
src/app.js:import 'angular-gettext/dist/angular-gettext.js';
```

PR #40, opened 2020-01-30
Test on fedora-coreos

martinpitt:


marusak:
I have two questions/uncertainties:

In commit message you explain that:
> Fedora CoreOS does not have Python 2 at all, and Python 3 is not
installed either.

But then you install python-3:
```
$(shell [ "$(TEST_OS)" = fedora-coreos ] && echo "--run-command 'rpm-ostree install python3'
```

When you do:
```
podman pull nginx
```
can docker use that image as well? (since you do `$(which podman 2>/dev/null || which docker) run nginx`)

martinpitt:
> But then you install python-3:

Argh, that was a leftover from rebasing, I'll remove it.

> can docker use that image as well

No, it totally can't, thanks for spotting. How *does* that work?? Turns out that RHEL Atomic host 7.7 actually has podman installed, so let's just use that.

martinpitt:
[This flake](https://209.132.184.41:8493/logs/pull-40-20200203-143838-7eadaa4f-fedora-coreos/log.html#4-2) is really annoying. I think I finally understand it now, it's rather hard to reproduce locally. I added another commit, let's see how it goes.

martinpitt:
@marusak: yaaay, green! can you PTAL?

PR #44, opened 2020-01-31
Various small cleanups

martinpitt:


martinpitt:
`python -m SimpleHTTPServer 1234` seems to fail without the `provision` on rhel-atomic, bah. (Works fine on fedora-coreos)

PR #45, opened 2020-01-31
Set page status for availability of an update

martinpitt:
Model this after cockpit-packagekit. ostree sends "change" events in
quick succession, and the first replies often don't include an available
update. To avoid flickering between "up to date" and "update available",
debounce the handler for 1s.

Rework the manifest to use a simple /updates path, and preload our page.

Factorize out the releaseName() filter to a separate function, so that
we can re-use it directly from JS.

The test currently only works on fedora-coreos, as RHEL and CentOS
Atomic have a too old cockpit that doesn't support page statuses yet.

 - [x] Add testing on fedora-coreos: PR #40

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@patternfly/react-core": "^3.153.3",
     "moment": "2.27.0",
     "react": "16.13.1",
-    "react-dom": "16.13.1"
+    "react-dom": "16.13.1",
+    "throttle-debounce": "2.2.1"
   }
 }
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -37,6 +37,7 @@ import { Spinner,
          DataToolbar, DataToolbarItem, DataToolbarContent,
        } from '@patternfly/react-core/dist/esm/experimental';
 import moment from 'moment';
+import { debounce } from 'throttle-debounce';
 
 import cockpit from 'cockpit';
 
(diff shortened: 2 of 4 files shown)


martinpitt:
Screenshots:

uptodate

available

martinpitt:
I haven't fully fixed testing on FCOS yet (PR #40), but that new test should already work. I'll trigger it here, so far I only ran it locally (and it succeeds there)

martinpitt:
> should already work

Ah no, it [fails](https://209.132.184.41:8493/logs/pull-45-20200131-143030-3f462365-fedora-coreos/log.html) due to missing `atomic` command. Seems I really need to get the coreos tests working properly first (but I need to understand in more detail what the tests are trying to do)

martinpitt:
Rebased on top of PR #40, and blocking on it. Now tests ought to succeed.

martinpitt:
I spent about way too long trying to debug that weird uncaught promise rejection crash already. This is somewhere deep in the guts between cockpit.transport and Angular.. needswork for now.

martinpitt:
Hm, I wasted so much time on this already.. This seems to be a really tricky issue between our page preloading and Angular not dealing well with page modifications. When cockpit's pkg/shell/base_index.js `lookup()` function changes the frame's `src=` attribute to update the hash, the Angular page ends up with a `{ state: 2 }` object in its `checkQueue`, which `processChecks` misinterprets as exception and prints it (in node_modules/angular/angular.js).

I really wonder if it's actually worth wasting more time on that. This is an ancient Angular version, not maintained any more, and it's also hard to understand our code. That time may be better spent with rewriting the page in React, it's not really that much code.

martinpitt:
The React rewrite in PR #105 landed :tada: so this is unblocked.

martinpitt:
This is finally unblocked, now that we landed the React rewrite. I updated this to the new code.

PR #48, opened 2020-02-04
Sync common CSS with cockpit

### LINKED ISSUE #47: Update lib/* from Cockpit, notably PF3 overrides and new RedHatText/RedHatDisplay fonts
Cockpit-ostree is out of date with the look of Cockpit. It needs the latest version of the PF3 overrides (to look like PF4) and the new RedHat* fonts (instead of Overpass).

martinpitt:
Fixes #47

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "7.8.3",
+    "@patternfly/react-styles": "3.6.26",
     "angular": "1.7.9",
     "angular-bootstrap-npm": "0.13.4",
     "angular-gettext": "2.4.1",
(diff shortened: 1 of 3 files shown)


martinpitt:
This is current master:

ostree-master

martinpitt:
This PR is a first attempt that syncs all shared files. It looks totally broken:

full-sync

So I guess I'll do that in smaller steps, and start with updating the fonts etc.

martinpitt:
I dialed this back to just syncing page.css and variables.less, and reverted the table/listing CSS changes. Now it looks quite close to master again:

page css

The only difference that I see is that the light-gray dashed lined box around "Currently using:" is gone. @garrett, I suppose that's not intended? But the font is still Open Sans. Not sure yet why, ostree just pulls in cockpit's base1/patternfly.css, and we set it there.

martinpitt:
Let's not do that. I'll rather look into a clean React+PF4 rewrite.

PR #51, opened 2020-02-06
Drop angular from src/utils.js

martinpitt:
This is preparation for rewriting cockpit-ostree in React.


martinpitt:
PR #50 works and includes these, nevermind.

PR #53, opened 2020-02-10
Drop less markers from *.css imports

martinpitt:
less ≥ 3.11 will try to look for a <filename>.less (e. g.
"page.css.less") which does not exist.


martinpitt:
Same fix as in Cockpit this morning: https://github.com/cockpit-project/cockpit/commit/65dfb4fc177901ec8

I visually tested this, and it looks alright.

PR #54, opened 2020-02-12
Confine Angular usage to app.js, separate ostree remote logic

martinpitt:
This is another prerequisite for porting to React.

### DIFF
--- a/src/remotes.js
+++ b/src/remotes.js
@@ -31,9 +31,7 @@ angular.module('ostree.remotes', [
 ])
 
 .factory("remoteActions", [
-    "$q",
-    "$timeout",
-    function($q, $timeout) {
+    function() {
         function listRemotes() {
             return cockpit.spawn(["ostree", "remote", "list"],
                                  { superuser: "try", err: "message" })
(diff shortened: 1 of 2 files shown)


martinpitt:
@KKoukiou : Friendly ping?

src/remotes.js
```diff
@@ -97,48 +95,29 @@ angular.module('ostree.remotes', [
...
-                    d.reject(ex);
-                })
-                .always(function () {
-                    file.close();
-                });
+            const file = getRemoteSettingsFile(name);
+            const section = getSectionName(name);
 
-            return d.promise;
+            const promise = file.modify(content => changeData(content, section, options));
```
KKoukiou: Did you miss the fail handler here?

KKoukiou: Ah actually it's trivial, not doing anything - sorry for noise.

martinpitt: Right, the original code just passed it on and just re-wrapped the rejection into a new defer.reject(). I simplified the code to simply pass on the original promise instead, and just kept the .always(file.close) cleanup.

martinpitt: Please do bring up the "noise" -- that's the point of reviews, to double-check :-) Thanks!

PR #90, opened 2020-05-24
Check for NPM dependency updates

cockpituous:
Check for any outstanding update of point version of any outstanding NPM package dependency
 * [ ] FAIL: npm-update

PR #91, opened 2020-05-26
test: Allow authorize messages

martinpitt:
This ignores the sudo journal messages, and thus fixes running against
Cockpit 219.

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -356,6 +356,7 @@ class OstreeRestartCase(MachineCase):
         b.wait_not_present(sel_nonexist)
 
         self.allow_restart_journal_messages()
+        self.allow_authorize_journal_messages()
 
     def testRebase(self):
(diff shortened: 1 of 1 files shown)


martinpitt:
This blocks at least https://github.com/cockpit-project/bots/pull/895 , https://github.com/cockpit-project/cockpit-ostree/pull/90 , and https://github.com/cockpit-project/bots/pull/895

martinpitt:
Meh, this now breaks [continuous-atomic](https://logs.cockpit-project.org/logs/pull-91-20200526-060928-3f54194f-continuous-atomic/log.html)

martinpitt:
Reverted to test API 211, as 219 also made all the tests appear as failed -- they don't produce proper TAP output right now. And I don't feel like debugging these two other tests right now, this is blocking too many things and thus is urgent.

PR #92, opened 2020-05-28
Build rpm for continuous-atomic in centos-7

martinpitt:
Binary rpms (even `noarch`) built on Fedora ≥ 31 cannot be installed any
more into continuous-atomic's rpm-ostree, due to an RPM format change.

When testing continuous-atomic, build the srpm inside a centos-7 image.
Add a test/rpmbuild-vm script for that.

For symmetry, move the call for local rpm builds out into
test/rpmbuild-local, and use the srpm as well instead of the .spec.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,6 @@ NODE_MODULES_TEST=package-lock.json
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/index.html
 
-ifeq ($(TEST_OS), rhel-atomic)
-MACRO_RHEL = --define "rhel 7"
-endif
-
 all: $(WEBPACK_TEST)
 
 #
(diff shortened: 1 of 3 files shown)


martinpitt:
With this, testing on continuous-atomic will work again locally, and this also finally unblocks updating our cockpit/tasks container from Fedora 30 (which is EOL now).

Makefile
```diff
@@ -119,21 +115,11 @@ srpm: dist-gzip $(PACKAGE_NAME).spec
...
-	  $(MACRO_RHEL) \
-	  $(PACKAGE_NAME).spec
-	find `pwd`/output -name '*.rpm' -printf '%f\n' -exec mv {} . \;
-	rm -r "`pwd`/rpmbuild"
-	rm -r "`pwd`/output" "`pwd`/build"
+# this is a noarch build, so local rpm build works fine mostly; the exception
+# is continuous-atomic, which does not get along with rpms build on Fedora ≥ 31
+rpm: srpm
+	srpm=`ls *.src.rpm | head -n1` && \
+	[ "$$TEST_OS" = continuous-atomic ] && test/rpmbuild-vm "$$srpm" centos-7 || test/rpmbuild-local "$$srpm"
```
KKoukiou, line 122: If `centos-7` image does not exist this rpmbuild-vm command will fail and this will end up in local rpm-build which had the known-problem with continuous atomic.

martinpitt: Thanks for spotting! Fixed.

PR #96, opened 2020-06-05
Stop importing cockpit's base1/patternfly.css

KKoukiou:
This is deprecated API and will be dropped at some point, in favor
of projects shipping their own CSS.

Follow https://github.com/cockpit-project/starter-kit/blob/master/webpack.config.js
approach on how to use PF CSS.

 - [x] Fix build failure: #97


martinpitt:
Argh, seems we need to do the rpm build inside the VM for rhel-atomic as well now:

   cockpit-ostree-176-2.el7.noarch
   error: Unrecognized archive format

See [rhel-atomic failure](https://logs.cockpit-project.org/logs/pull-96-20200605-161045-f53b9a60-rhel-atomic/log.html)

martinpitt:
Fixed the build failure #97

garrett:
I've looked into this a bit... it seems that `ostree.css` sets the font to sans-serif in the `body` block... but I can't figure out where that's coming from. (This is one of those times it would really help to have working source maps.) I guess it's from some include from node_modules as I can't seem to find it grepping through the source.

But even worse, there are no PF4 variables for the RH fonts... and RedHatText is nowhere to be found. I think both are because we're not using PF4?

Would it make sense to pull in PF4, but only include the bare minimum — perhaps just enough to get the fonts?

garrett:
Fonts look correct.

PR #97, opened 2020-06-06
Build rpm for rhel-atomic in centos-7

martinpitt:
RPMs built in the tasks container now fail on rhel-atomic as well, so
build it in the VM.


martinpitt:
Fixes https://logs.cockpit-project.org/logs/pull-96-20200605-161045-f53b9a60-rhel-atomic/log.html

PR #105, opened 2020-06-23
React rewrite

KKoukiou:


andreasn:
Took this for a quick spin. Some initial thoughts:

Screenshot from 2020-06-23 14-51-58

* The content at the top does not span the entire width of the window. Our other pages behaves mostly like this, so I think this behaviour should apply here too.
* _Repository_, _local_, and _Branch_ have the exact same spacing between them, making it hard to tell what labels belong together. Best is probably to add some more spacing between _local_ and _Branch_
* It would be nice to have the _Check for updates_ button at the very right. This would be consistent to what we do on other pages (overview, logs, (packagekit) software updates)
* There is a line between the section `pf-c-page__main-nav` and `available-deployments`, that I don't think adds any function.
* The Running/Available labels feels like they float off a little where they are now. Needs to be moved closer to the branch name, possibly swap places. Not sure what's best.

andreasn:
And also the branch switcher dialog:

Screenshot from 2020-06-23 15-15-10
Screenshot from 2020-06-23 15-15-24

* I think the rows could benefit with some lines top and bottom. Helps connect the branch name label to the Edit icon, and makes the expanded section for editing and adding a new branch easier to separate.
*  I liked the _Apply/Add_ button better when it was in the right corner of the expanded sections. Helped it stand out from the buttons of the dialog itself better.
* The _Name_ and _URL_ text fields doesn't span the entire width of it's section. This is bad when you have URL around the length of `https://kojipkgs.fedoraproject.org/compose/ostree/repo/`
* _Name_ and _URL_ has a big gap between them and the corresponding text field. I think we should move to left aligned labels at some point, since that's the Patternfly default, but probably all of Cockpit at once if so.

KKoukiou:
> Took this for a quick spin. Some initial thoughts:
> 
> Screenshot from 2020-06-23 14-51-58
> 
>     * The content at the top does not span the entire width of the window. Our other pages behaves mostly like this, so I think this behaviour should apply here too.

I actually don't have any custom CSS on this - I am just using the Toolbar component from PF4 https://www.patternfly.org/v4/documentation/react/components/toolbar

I would prefer to not change the default design from PF4 if we don't get any benefit from the changes.
> 
>     * _Repository_, _local_, and _Branch_ have the exact same spacing between them, making it hard to tell what labels belong together. Best is probably to add some more spacing between _local_ and _Branch_

Same here, it's just the Toolbar component's default spacing between labels.
> 
>     * It would be nice to have the _Check for updates_ button at the very right. This would be consistent to what we do on other pages (overview, logs, (packagekit) software updates)

Pasting for comparison here the screenshot of the packagekit page.
Screen Shot 2020-06-23 at 15 40 44

The difference is that that one does not have a toolbar, it's having some custom page header.

The advantage of reusing the Toolbar as is, is that the mobile mode is fully controlled by PF4 now.

> 
>     * There is a line between the section `pf-c-page__main-nav` and `available-deployments`, that I don't think adds any function.

Do you prefer it without? Check the diff before-after.

Screen Shot 2020-06-23 at 15 35 19

Screen Shot 2020-06-23 at 15 35 02

> 
>     * The Running/Available labels feels like they float off a little where they are now. Needs to be moved closer to the branch name, possibly swap places. Not sure what's best.

KKoukiou:
> And also the branch switcher dialog:
> 
> Screenshot from 2020-06-23 15-15-10
> Screenshot from 2020-06-23 15-15-24
> 
>     * I think the rows could benefit with some lines top and bottom. Helps connect the branch name label to the Edit icon, and makes the expanded section for editing and adding a new branch easier to separate.

Again this is a component from PF4 http://127.0.0.2:9091/updates-react
It does not have borders, I would prefer to not adjust the original PF4 design. We can always create an issue to PF4 is you feel that their design is suboptimal.
Or use another component. What do you think?

> 
>     * I liked the _Apply/Add_ button better when it was in the right corner of the expanded sections. Helped it stand out from the buttons of the dialog itself better.

The PF4 forms have the buttons on the bottom. https://www.patternfly.org/v4/documentation/react/components/form
I think it makes sense - especially for keyboard navigation because when one completes the form the next tab shoud be the buttons where they should apply the form.

I don't have really strong opinion on this, but as I have said multiple time I really like to follow blindly PF4 ;) 
That said I don't mind switching to the old design, but let's consider the keyboard navigation issue I pasted above.

> 
>     * The _Name_ and _URL_ text fields doesn't span the entire width of it's section. This is bad when you have URL around the length of `https://kojipkgs.fedoraproject.org/compose/ostree/repo/`

I will add a min-width to it.

> 
>     * _Name_ and _URL_ has a big gap between them and the corresponding text field. I think we should move to left aligned labels at some point, since that's the Patternfly default, but probably all of Cockpit at once if so.

The spacing is the default form spacing from PF4.

Also the PF4 default is not horizonal form. But I don't have a preference, I can change it.
https://www.patternfly.org/v4/documentation/react/components/form

martinpitt:
Force-pushed to fix merge conflict with recently merged moment npm update.

martinpitt:
@KKoukiou : build fails with
```
ERROR in ./src/app.jsx
Module not found: Error: Can't resolve './changeRemoteModal.jsx' in '/home/user/cockpit-project/bots/make-checkout-workdir/src'
 @ ./src/app.jsx 52:0-60 631:89-106
```
It seems your last force-push moved it out of app.jsx into a new file. git pilot error? (forgot `git add`)

src/app.jsx
```diff
@@ -0,0 +1,530 @@
...
+    return (
+        <DataListItem aria-labelledby={id}>
+            <DataListItemRow>
+                <DataListItemCells dataListCells={[
+                    <DataListCell key="name" width={4}> <span className="deployment-name" id={id}>{name}</span> </DataListCell>,
+                    <DataListCell key="state" width={4}><span className="deployment-status">{state}</span></DataListCell>,
+                    <DataListCell key="action" width={2}>
+                        {isUpdate(info) && <Button variant="secondary"
+                                                   onClick={() => doUpgrade(info.osname.v, info.checksum.v)}
+                                                   isDisabled={!!client.local_running}>{_("Update Reboot")}</Button>}
```
martinpitt, line 317: Should be "Update *and* Reboot"

src/changeRemoteModal.jsx
```diff
@@ -0,0 +1,261 @@
...
+
+    return (
+        <Form isHorizontal>
+            <Title headingLevel="h3" size="l">
+                {_("Add New Repository")}
+            </Title>
+            {addNewRepoError && <Alert variant="danger" isInline title={addNewRepoError} />}
+            <FormGroup label={_("Name")}
+                       fieldId="new-remote-name"
+                       helperTextInvalid={_(_("Please provide a valid name"))}
```
martinpitt, line 160: Let's not translate it *that* hard! :grin:

src/changeRemoteModal.jsx
```diff
@@ -0,0 +1,261 @@
...
+                       isRequired>
+                <TextInput id="new-remote-name"
+                           value={newRepoName}
+                           isRequired
+                           type="text"
+                           onChange={name => setNewRepoName(name)} />
+            </FormGroup>
+            <FormGroup label={_("URL")}
+                       fieldId="new-remote-url"
+                       helperTextInvalid={_(_("Please provide a valid URL"))}
```
martinpitt, line 171: Same i36n

src/changeRemoteModal.jsx
```diff
@@ -0,0 +1,261 @@
...
+                          id="gpg-verify"
+                          isChecked={isTrusted}
+                          onChange={(checked, ev) => {
+                              setIsTrusted(!isTrusted);
+                          }} />
+            </FormGroup>
+            <FormGroup fieldId="add-another-key">
+                {!addAnotherKey ? <Button isInline variant="secondary" id='add-another-key' onClick={() => setAddAnotherKey(true)}>{_("Add Another Key")}</Button>
+                 : <TextArea id='gpg-data'
+                             placeholder="Begins with '-----BEGIN GPG PUBLIC KEY BLOCK-----'"
```
martinpitt, line 246: This on the other hand needs *more* i18n; ideally without the `---BEGIN` stuff. Perhaps `_("Begins with '$0'`")` and cockpit.format?

martinpitt:
I fixed the two remaining failures in `OstreeRestartCase.testOstree` and the four little issues above. The remaining TODO is now to get rid of the HACK commit.

martinpitt:
There seems to be [one failure on continuous-atomic](https://logs.cockpit-project.org/logs/pull-105-20200626-084532-a84b50b7-continuous-atomic/log.html) which we need to look into, otherwise tests look good.

Moreover, it seems the root cause for the HACK dissolved into nothingness :tada:  . It works just fine without, so let's drop it.

martinpitt:
Pushed trivial test case fix, it just needed to drop the continuous-atomic special case. Thanks @KKoukiou, that was a really great piece of work!

PR #109, opened 2020-06-29
Merge changes from starter-kit up to f2a11ba1dee

martinpitt:
This gives us several nice improvements, like
    
 - using run-test for TAP output and auto-retries,
 - `make` not rebuilding stuff unnecessarily, or
 - more eslint checks.

But mostly it keeps our build system as much in sync as possible.


martinpitt:
Hmm, this is represented as 130 commits. This is nice for making future `git merge`s easier, but is a lot of noise.

marusak:
Hmm, 130 commits is a lot of noise. Could maybe just the package.json refreshes squashed into one?

martinpitt:
I played around with `git rebase -i --rebase-merges origin/master` and such, but if I even just squash a few commits together, it becomes totally broken. With plain rebases/squashing, git loses the history of which commits already got merged, so the next time it's going to end up with all these conflicts again. But I agree, it's probably easier to just do this manually, and note down the ref of starter-kit that was used for merging (before squashing), and then the next time we just manually apply the diff from starter-kit from the given ref to master. Kind of low-tech and uncool, but avoids all that noise.

martinpitt:
Ugh, the net diff was zero (proven by GitHub's interdiff link), but this now fails with
```
/home/user/cockpit-project/bots/make-checkout-workdir/src/app.jsx
  28:8  error  Expected a line break after this opening brace  object-curly-newline
  36:8  error  Expected a line break after this opening brace  object-curly-newline
```

Not sure what happened, but these are legit. I added two separate commits to fix them.

PR #114, opened 2020-07-12
client: Convert some jQuery to JS Promises

martinpitt:
Keep cockpit.defer() for run_transaction() as that uses progress
notify.
   
connect() will be ported separately as that is very complex.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -148,7 +148,7 @@ function Packages(promise, transform) {
         .catch(ex => {
             this.error = cockpit.message(ex);
         })
-        .always(() => {
+        .finally(() => {
             this.ready = true;
             this.dispatchEvent("changed");
         });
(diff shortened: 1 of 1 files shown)


martinpitt:
Meh, I don't understand the [continuous-atomic failure](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-114-20200721-125905-f830a42e-continuous-atomic/log.html#5-2), and I don't feel like tracking this down now. This is really not important, so let's just do that small cleanup for now.

martinpitt:
@KKoukiou : Do you have some time to review this in the next days? Thanks!

src/client.js
```diff
@@ -700,9 +696,11 @@ class RPMOSTreeDBusClient {
                                     }
                                 });
                             transaction_client.call("/", TRANSACTION, "Start");
```
KKoukiou, line 698: Failures from this call won't be caught in the fail unless this is returned.

martinpitt: Right, the code doesn't change that. Perhaps we should do that though, good catch!

PR #124, opened 2020-08-04
Bump test API to 224

martinpitt:


martinpitt:
This was mostly to validate the new centos-7 image from https://github.com/cockpit-project/bots/pull/1118 , which I accidentally landed. Seems all is well.

PR #139, opened 2020-09-18
test: Bump test/common to 228

marusak:
On the login page the `Reuse password` checkbox was dropped and that
makes all the tests to fail.
See https://github.com/cockpit-project/cockpit/commit/ef97e7e9a28a

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -41,6 +41,12 @@ CHECKOUT_LOCATION = "/var/local-tree"
 RPM_LOCATION = "/usr/share/rpm"
 KEY_ID = "95A8BA1754D0E95E2B3A98A7EE15015654780CBD"
 
+# old cockpit/ws container on rhel-atomic still has "[ ] Reuse my password for privileged tasks"
+login_opts = {}
+if os.environ.get("TEST_OS") == "rhel-atomic":
+    login_opts["legacy_authorized"] = True
+
+
 def get_list_item(index):
     # there are more <li> inside the item tables, so only count direct children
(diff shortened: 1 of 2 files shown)


marusak:
rhel-atomic is using some prehistoric version of cockpit. It will probably need `legacy_authorized` in `login_and_go()`.

martinpitt:
Indeed, I pushed a fix.

martinpitt:
Still breaks [page status on fedora-coreos](https://logs.cockpit-project.org/logs/pull-139-20200921-204636-f0041ad3-fedora-coreos/log.html)

marusak:
Blocks https://github.com/cockpit-project/bots/pull/1246

test/check-ostree
```diff
@@ -659,7 +668,8 @@ class OstreeCase(MachineCase):
         rhsmcertd_hack(m)
 
         # preloading works, no updates available
-        self.login_and_go("/system")
```
marusak, line 662: This seems like copy-pasta problem. Why it now goes to `/updates` when it used to go to `/system`? And it checks `Health` card so it seems incorrect.

test/check-ostree
```diff
@@ -685,7 +695,7 @@ class OstreeCase(MachineCase):
 
         # check with new session
         b.logout()
-        self.login_and_go("/system")
```
marusak, line 688: the same problem.

marusak:
> Still breaks page status on fedora-coreos

It seems that you changed page to which we log into. I'll fix it.

martinpitt (approved):
Haha, of course, copy pasta! Dakujem!

PR #148, opened 2020-10-01
test: Bump test/common to b7fc

Gundersanne:
See https://bugzilla.redhat.com/show_bug.cgi?id=1884236.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,7 @@ bots:
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
-	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 228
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -173,7 +173,7 @@ bots:
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
-	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 228
+	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git b7fc7bd5e517659c45531dcf84505902a17ec9e8
```
martinpitt, line 176: That's a little wild, but indeed on second look nicer than ading it to the test's setUp() temporarily. :-)

martinpitt (approved):
Want to test this against the new image with the `@bots#123` syntax?

Makefile
Gundersanne: yea and releasing 229.1 seems like a bit much :p

PR #153, opened 2020-10-10
cockpituous: Run release in a GitHub workflow

martinpitt:
Enter the new world of GitHub actions [1]/GitLab pipelines [2]. This
simplifies our end of the infrastructure considerably:

 * No need any more to set up webhooks, all the relevant configuration
   is right in the workflow file.

 * Does not need any infrastructure on our side any more, and thus works
   for third-party projects. They just need to set up their own secrets.

 * Gets rid of our webhook's special case for releases, which spawns an
   OpenShift Job (which means we can replace that webhook code with
   something more generic). Structually, spawning a Job is exactly like
   running an action/pipeline.

 * GitHub automatically provides a temporary `GITHUB_TOKEN` with
   sufficient write access to the project to publish a release, so we
   don't need to carry around that secret.

[1] https://github.com/features/actions
[2] https://docs.gitlab.com/ee/ci/pipelines/

 - [ ] needs some adjustments to our release container: https://github.com/cockpit-project/cockpituous/pull/343
 - [x] Disable release (tag) webhook for cockpit-ostree

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: release
+on:
+  push:
+    tags:
+      # this is a glob, not a regexp
+      - '[0-9]*'
+jobs:
+  cockpituous:
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/cockpit/release
+    steps:
+      - name: Set up configuration and secrets
+        run: |
+          # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+
+          echo '${{ secrets.SSH_KNOWN_HOSTS }}' > ~/.ssh/known_hosts
+          echo '${{ secrets.FEDPKG_SSH_PUBLIC }}' > ~/.ssh/id_rsa.pub
+          echo '${{ secrets.FEDPKG_SSH_PRIVATE }}' > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          echo 'cockpit' > ~/.config/bodhi-user
+          echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/github-token
+          echo '${{ secrets.COPR_TOKEN }}' > ~/.config/copr
+          echo '${{ secrets.COCKPIT_FEDORA_PASSWORD }}' > ~/.fedora-password
+
+      - name: Run cockpituous
+        run: |
+          # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+          cd /build
+          release-runner -r https://github.com/$GITHUB_REPOSITORY -t $(basename $GITHUB_REF) ./cockpituous-release
--- a/README.md
+++ b/README.md
@@ -96,6 +96,10 @@ starter-kit includes an example [cockpitous release script](./cockpituous-releas
 that builds an upstream release tarball and source RPM. Please see the above
 cockpituous documentation for details.
 
+Automatic releases are triggered through the [release.yml](.github/workflows/release.yml)
+[GitHub action](https://github.com/features/actions) workflow. This uses the
+[cockpit-project organization secrets](https://github.com/organizations/cockpit-project/settings/secrets).
+
 # Further reading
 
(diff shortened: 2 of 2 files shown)


martinpitt:
I tested this with a [slightly modified workflow](https://github.com/martinpitt/cockpit-ostree/commit/5b3055a64393f77d53c18909a2e1602ecc382059) that does *not* install the COPR token (I wanted the "pre-commit" phase to fail as late as possible), and [disabling most Fedora releases](https://github.com/martinpitt/cockpit-ostree/commit/2352e553bf6df8c4a76bf897bfb7d43c2bf10c91). That was as far as I dared to go on my fork, without actually uploading something to Fedora. It created a [release draft](https://github.com/martinpitt/cockpit-ostree/releases/tag/untagged-b904d03a1d9cf04d7b08) and a [Fedora scratch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=53143983). The [whole run/log](https://github.com/martinpitt/cockpit-ostree/runs/1235328525?check_suite_focus=true).

For that test run I uplodaed the secrets to [my forked project](https://github.com/martinpitt/cockpit-ostree/settings/secrets). For production we want to upload them to the [cockpit organization](https://github.com/organizations/cockpit-project/settings/secrets) instead, so that all our projects can use them.

I put the release secrets into a private gitlab CEE repo that me, @Gundersanne, and @marusak can access (please ping me for the URL, but it's right next to our "infra" repo).

martinpitt:
I'd like to do a release of cockpit-ostree soon with this, after landing this PR. If something breaks, I'll send follow-up fixes,  of course. Fortunately our release process is reasonably robust.

@Gundersanne, @marusak, WDYT about this approach?

.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
+name: release
+on:
+  push:
+    tags:
+      - '[0-9]*'
```
marusak, line 5: what about point releases? `12.1` would not match this regex.

.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
...
+        run: |
+          # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+
+          echo '${{ secrets.SSH_KNOWN_HOSTS }}' > ~/.ssh/known_hosts
+          echo '${{ secrets.FEDPKG_SSH_PUBLIC }}' > ~/.ssh/id_rsa.pub
+          echo '${{ secrets.FEDPKG_SSH_PRIVATE }}' > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          echo 'cockpit' > ~/.config/bodhi-user
+          echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/github-token
```
marusak, line 22: what are these 'secrets'? Where does it come from?

marusak (changes requested):
Looks super simple and if that 'just works'™ that wow!
I think that regex should also count with point releases (not likely we need it here, but once we copy this to other projects) and have one question.

.github/workflows/release.yml
martinpitt: This is not a regexp, it's a [glob](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags). Thus this just means "tag must start with a number".

.github/workflows/release.yml
martinpitt: See the [comment above](https://github.com/cockpit-project/cockpit-ostree/pull/153#issuecomment-706545726). You should be able to see them. This still needs documenting in the README of course, I'll send a PR once everything works.

.github/workflows/release.yml
martinpitt: I added a comment about this.

.github/workflows/release.yml
martinpitt: I added a README stanza to explain how the release is triggered and where the secrets live.

marusak (approved):
Nice, thanks!

martinpitt:
OK, let's take the plunge -- I disabled the tag creation webhook.

martinpitt:
Nice, it all worked!

PR #154, opened 2020-10-12
Run npm-update in a GitHub workflow

martinpitt:
This avoids a lot of moving parts in our infrastructure (webhook,
npm-trigger roundtrip, tasks container), works in exactly the same way
for independent third-party projects, and does not need *any* secret
other than the automatically provided GitHub token.

Let this run early Tuesday morning, so that we have enough time to
review/land/fix the updates before our Wednesday release. Also add a
manual trigger, so that we get a button to run it on demand.

 - [x] Fix bots to get along with restricted token: https://github.com/cockpit-project/bots/pull/1315


martinpitt:
I developed this on my fork, where [this workflow run](https://github.com/martinpitt/cockpit-ostree/runs/1240711640?check_suite_focus=true) created [this PR](https://github.com/martinpitt/cockpit-ostree/pull/3).

.github/workflows/npm-update.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Run bot
+        run: |
+          make bots
+          # we do the PR ourselves, much easier and avoids the need for a custom token
+          bots/npm-update --verbose --dry
+          # compute the changed module from git diff; that's a bit awkward, let's perhaps add an npm-update option?
```
Gundersanne, line 27: agreed

Gundersanne (approved):
lgtm!

.github/workflows/npm-update.yml
```diff
@@ -0,0 +1,39 @@
...
+          git diff package.json | sed -n '/^+ / { s/^[^"]*"//; s/".*$//; p; q }' >> $GITHUB_ENV
+
+      # this silently skips creating the PR if there are no changes, and updates an existing PR
+      - name: Create/update Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: "package.json: Update ${{ env.MODULE }} package dependency"
+          title: "[bot] package.json: Update ${{ env.MODULE }} package dependency"
+          branch: bots/npm-update
+          delete-branch: true
```
marusak, line 38: The update happens every Tuesday. If we don't close/merge PR in one week, this will become problematic. can we just use generated branch name (at least `npm-update-<date>`)?

.github/workflows/npm-update.yml
marusak: yeah, this is ugly

marusak (changes requested):

.github/workflows/npm-update.yml
martinpitt: Problematic how? I specifically chose to *not* pile up lots of branch names. The workflow will update an already existing PR in the following week if it is still open. I find that much nicer than having potentially duplicated/outdated PRs.

marusak: You start with master, create commit and then try to push into branch that already has commit in it. It may cause conflicts (and likely will as they might update the same file). Or am I misunderstanding what this does?

martinpitt: According to the [documentation](https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#how-the-action-works), it force-pushes on changes, and it always starts with the base branch (current master).

If we want to, we can switch to a [timestamp suffix](https://github.com/marketplace/actions/create-pull-request#alternative-strategy---always-create-a-new-pull-request-branch) just like we have with cockpituous now, if we don't like the behaviour. The docs explicitly don't recommend that, though. But I'd actually like to try and see how the default behaviour works for us, and if we don't like it, we can still change the behaviour?

Gundersanne: I'd be ok then with seeing the default behaviour, especially if it's recommended in the docs.

martinpitt:
Moving to draft. There is still some impedance mismatch between what the bot is trying to do, and separately creating a single PR. I'm leaning towards adding a `--no-fork` option to tasks to create branches right in the originating repo (which isn't a big deal with auto-cleanup), but otherwise let the bots create the PRs. If we want to throw out our code for that and replace it with standard actions, we can still do that later. We'd still get the "no custom secrets", "effortlessly ports to third-party projects", and "needs no own infra" gains, which are the important ones.

martinpitt:
> I'm leaning towards adding a --no-fork option to tasks to create branches right in the originating repo

Turns out we don't need that -- our bots `Task` already DTRT mostly ("We actually own the origin repo, so use that."). In the github action case, it will get a token for the original repo, not for the `cockpituous` user/fork. It only [falls over](https://github.com/martinpitt/cockpit-ostree/runs/1260681019?check_suite_focus=true) because the token can't access the /user API.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/1315 to make bots get along with restricted tokens. I pushed the commit from this PR plus [a hack to use my bots fork for #1315](https://github.com/martinpitt/cockpit-ostree/commit/d75b1cef90e3e2de9) to my fork's master branch, [ran the action](https://github.com/martinpitt/cockpit-ostree/runs/1263315300?check_suite_focus=true), and it [produced a valid PR](https://github.com/martinpitt/cockpit-ostree/pull/9) similar to what cockpituous does.

I really want to deprecate these cockpituous forks. It takes some work to set up and distribute the token, and it [keeps piling up thousands of branches](https://github.com/cockpituous/cockpit/branches/active) which never get cleaned up. As GitHub still doesn't reliably clean up branches on merging, I'll add another action to do that for branches on the main repo, i.e. those produced by bots (normal devs usually have the branches on their forked repo).

@marusak, @Gundersanne : WDYT?

martinpitt:
> As GitHub still doesn't reliably clean up branches on merging

Ah, I don't have the "automatically delete head branches" option off on my fork, but it's on for c-ostree. I'd say, let's go through a complete bot PR, and if it doesn't get cleaned up properly, I'll look into an explicit cleanup action.

marusak (approved):
Looks good, I'll now take a look at https://github.com/cockpit-project/bots/pull/1315 to unblock. Thanks!

martinpitt:
Yay, let's do this then :) I'll check this Tuesday morning, when the action should run automatically.

PR #158, opened 2020-10-20
actions: npm-update adjustments

martinpitt:
- Run the job daily, just like cockpituous does (roughly, it was much
   more probabilistic). Each run will only create one PR at most, so
   with weekly we would miss or drag out important updates.

 - Drop verbosity. It was useful for rolling this out, but it's prone to
   expose secrets, and not very interesting as a run is rather simple to
   reproduce locally.


marusak (approved):
Thanks! You are right that with one update per week we could be slow to adapt to new packages. If once per day would become too much we can dial it down later on.

PR #31, opened 2019-10-30
Enable copr builds for packit config

dhodovsk:
We enabled copr builds in the packit config for you.

After merging this PR, you are just a few steps away from RPM builds being automatically triggered on your PR's.
It means, that you'll be able to try and play with your change, packaged as an RPM.

What are the next steps?
* Install [Packit-as-a-Service github application](https://github.com/marketplace/packit-as-a-service) in your repo
* In case you are first user, wait for account approval
* Enjoy the built RPMs!

For more info, please:
 * check out the documentation: https://packit.dev/docs/
 * contact [@packit-service/the-packit-team](https://github.com/orgs/packit-service/teams/the-packit-team)

PR #160, opened 2020-11-26
test: Pull nginx image from quay.io

martinpitt:
In the Red Hat internal Cockpit CI we constantly run into docker.io pull
rate limits. Use the already existing quay.io mirror of that image from
the app-sre team.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -21,7 +21,8 @@ if [ ! -d /var/local-tree ]; then
     ostree checkout "$checkout" /var/local-tree
 fi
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Blocks https://github.com/cockpit-project/bots/pull/1411

martinpitt:
Uh, seems packit now automatically runs PRs -- I was still working on that in #31, but that's still blocked due to a packit bug. So ignore these failures for now.

marusak (approved):
Thanks

PR #166, opened 2021-01-05
Makefile: Output a path to the generated archive

marusak:
This is needed for packit to work.
Also reusing `$(TARFILE)` macro.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,7 @@ devel-install: $(WEBPACK_TEST)
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
 dist-gzip: $(TARFILE)
+	echo `pwd`/$(TARFILE)
 
 # when building a distribution tarball, call webpack with a 'production' environment
 # we don't ship node_modules for license and compactness reasons; we ship a
@@ -113,7 +114,7 @@ $(TARFILE): $(WEBPACK_TEST) $(PACKAGE_NAME).spec
 	mv node_modules node_modules.release
 	touch -r package.json $(NODE_MODULES_TEST)
 	touch dist/*
-	tar czf $(PACKAGE_NAME)-$(VERSION).tar.gz --transform 's,^,$(PACKAGE_NAME)/,' \
+	tar czf $(TARFILE) --transform 's,^,$(PACKAGE_NAME)/,' \
 		--exclude $(PACKAGE_NAME).spec.in \
 		$$(git ls-files) lib/patternfly/*.scss package-lock.json $(PACKAGE_NAME).spec dist/
 	mv node_modules.release node_modules
--- a/packit.yaml
+++ b/packit.yaml
@@ -1,7 +1,7 @@
 specfile_path: cockpit-ostree.spec
 synced_files:
   - cockpit-ostree.spec
-upstream_project_name: cockpit-ostree
+upstream_package_name: cockpit-ostree
 downstream_package_name: cockpit-ostree
 actions:
   post-upstream-clone: make cockpit-ostree.spec


marusak:
@martinpitt I thought I would take a look at this as I never really touched ostree nor packit :)

marusak:
Oh what?
```
Running command: make dist-gzip
Output of ['make', 'dist-gzip']:
# if it exists already, npm install won't update it; force that so that we always get up-to-date packages
rm -f package-lock.json
# unset NODE_ENV, skips devDependencies otherwise
env -u NODE_ENV npm install
npm WARN deprecated eslint-loader@3.0.4: This loader has been deprecated. Please use eslint-webpack-plugin
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated fsevents@2.1.3: Please update to v 2.2.x
npm WARN deprecated core-js@2.6.12: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.

Message: Preparing of the upstream to the SRPM build failed: The create-archive action did not output a path to the generated archive. Please make sure that you have valid path in the single line of the output.
Exception: PackitSRPMException('Preparing of the upstream to the SRPM build failed: The create-archive action did not output a path to the generated archive. Please make sure that you have valid path in the single line of the output.')
You can retrigger the build by adding a comment (`/packit copr-build`) into this pull request.
Please join the freenode IRC channel #packit for the latest info.
```

The output from `make dist-git` seems waay too short.

Makefile
```diff
@@ -103,6 +103,7 @@ devel-install: $(WEBPACK_TEST)
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
 dist-gzip: $(TARFILE)
+	echo $(TARFILE)
```
martinpitt, line 106: That certainly *looks* right according to https://packit.dev/docs/actions/ ..

marusak:
oh common, locally it already worked. (It needed path, not just file name, so `./` did the trick)

packit.yaml
```diff
@@ -5,4 +5,6 @@ upstream_project_name: cockpit-ostree
 downstream_package_name: cockpit-ostree
 actions:
   post-upstream-clone: make cockpit-ostree.spec
-  create-archive: make dist-gzip
+  create-archive:
+    - make dist-gzip
+    - make dist-gzip-dump
```
martinpitt, line 10: OOI, what was wrong with letting dist-gzip output the path by itself?

Makefile
```diff
@@ -104,6 +104,9 @@ devel-install: $(WEBPACK_TEST)
 
 dist-gzip: $(TARFILE)
 
+dist-gzip-dump:
+	echo ./$(TARFILE)
```
martinpitt, line 108: Not sure why we need this new target, instead of merging it with dist-gzip. But if we have it, then IMHO it should depend on `$(TARFILE)` by itself, so that you don't have to call two targets and run into race conditions with `-j`.

martinpitt (changes requested):

packit.yaml
marusak: Nothing really, I am still trying to figure out what is the problem. Locally it worked just fine without any additional target. Here it does not work. It outputs only a few lines from `make dist-gzip` so I thought maybe it does not see the last line, so I tried if having separate target would help. It did not, even though it is now visible in the output. Still playing around. Once I find out the solution I'll clean up after myself.

marusak:
Ok, I think it actually does not create the archive.

marusak:
I am now trying to build this in `usercont/sandcastle` container ( I saw this in https://github.com/packit/sandcastle/issues/85) and now it fails with:
```
ERROR: You need to install the 'sassc' package to build this project.
```
So either we need to install it or get it installed like we got `npm`? https://github.com/packit/sandcastle/pull/30

marusak:
Ok so indeed sassc needs to be installed into sandbox, sent https://github.com/packit/sandcastle/pull/90
Also not all logs are shown. Talked to packit team and they will take a look at it.
Until the PR is merged and deployed, this is blocked.

martinpitt:
@marusak: The changes here look good and harmless, and can be pre-validated by our CI. If you want to re-push without "no-test", I'm happy enough to land them already.

marusak:
@martinpitt tests were green. I added one more commit.
Locally `packit srpm` and `packit local-build` both work.

martinpitt (approved):
Danke!

PR #168, opened 2021-01-25
Use cockpit library components and webpack helpers from cockpit.git

martinpitt:
Drop the cockpit-po-plugin.js, sassc-loader.js, variables.scss etc.
code copies, and check them out from a stable cockpit tag, similarly to
how we already get the PatternFly CSS.

Also move them from lib/ to src/lib/ for consistency with starter-kit.

Install bootstrap-sass npm module, as the current variables.scss imports
it. It's not really necessary for cockpit-ostree, but scss compilation
would fail otherwise. (Let's try to drop it from cockpit's version.)

Ported from starter-kit commit bb064229967.


martinpitt:
I tested this visually locally, and it looks fine to me.

package.json
```diff
@@ -19,6 +19,7 @@
     "@babel/preset-react": "^7.0.0",
     "babel-eslint": "^10.0.3",
     "babel-loader": "^8.0.6",
+    "bootstrap-sass": "^3.4.1",
```
KKoukiou, line 22: I really wonder how it worked at the podman repo!??

KKoukiou (approved):

package.json
KKoukiou: Oh I know, we don't include there the form-layout.scss that imports the variables.scss file.

KKoukiou:
Btw usage of form-layout layout should be replaced with Form component from PF4. This file is from a pre PF4 era.

PR #172, opened 2021-02-25
packit: Fix archive creation

martinpitt:
Apply the same memory consumption reduction and tarfile printing fixes
as cockpit-podman recently did.

Also clean up `synced_files`, this should not be necessary any more with
current packit.


martinpitt:
Yay, no more red on ostree PRs :tada:

marusak (approved):
Thanks!

PR #173, opened 2021-03-23
pf4: Use SVG spinner to fix direction and glitches

garrett:
Blocked on:
* [x] #174 

Addresses https://github.com/cockpit-project/cockpit/issues/15573

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -111,7 +111,7 @@ const Curtain = ({ state, failure, message, reconnect }) => {
 
     let icon = null;
     if (state === 'connecting')
-        icon = <Spinner size="xl" />;
+        icon = <Spinner isSVG size="xl" />;
     else if (failure)
         icon = <EmptyStateIcon icon={ExclamationCircleIcon} />;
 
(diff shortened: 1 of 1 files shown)


garrett:
This doesn't seem to work, but it also doesn't seem to break things.

My guess is that the version of PatternFly we're using for cockpit-ostree is too old. Once that's upgraded, this PR should work.

KKoukiou:
@garrett this needs that you change the PF version in the package.json

garrett:
package.json uses the same version as Cockpit, however, the spinners are still going the wrong way and I can't find any additional use of a React `<Spinner>` in the code.

KKoukiou:
> package.json uses the same version as Cockpit, however, the spinners are still going the wrong way and I can't find any additional use of a React `<Spinner>` in the code.

We are interested in @patternfly/react-core and it's not the same version as in Cockpit. 
https://github.com/cockpit-project/cockpit-ostree/blob/master/package.json#L53

KKoukiou:
Oh I know, there was a breaking change release of course.

martinpitt:
> I am not sure why the npm-update script does not want to update patternfly here.

It only updates minor versions, not major ones.

martinpitt:
Unblocked, please rebase.

KKoukiou:
> Unblocked, please rebase.

It's actually fine, there are not conflicts here. I will just test rebase locally and test it.

src/app.jsx
```diff
@@ -190,7 +190,7 @@ const OriginSelector = ({ os, remotes, branches, branchLoadError, currentRemote,
                                 id="check-for-updates-btn"
                                 isDisabled={!!client.local_running || !!progressMsg}
                                 onClick={checkForUpgrades}>
-                            {!(!!client.local_running || !!progressMsg) ? _("Check for Updates") : <Spinner size="sm" />}
+                            {!(!!client.local_running || !!progressMsg) ? _("Check for Updates") : <Spinner isSVG size="sm" />}
```
KKoukiou, line 193: This should be changed to use isLoading property.

KKoukiou (changes requested):

KKoukiou (approved):

Issue #177, opened 2021-03-24
Weird spacing at top of the page


garrett:
1. The top bar has inconsistent padding. 
2. The body has a line for some odd reason and the padding also seems off.
3. The width of the page and the cards in the page feels strange.
 
![Screenshot_2021-03-24 Software Updates - garrett Rain(5)](https://user-images.githubusercontent.com/10246/112340680-11ca9d00-8cc1-11eb-9000-ccbe19fba4eb.png)

KKoukiou:
>     1. The top bar has inconsistent padding.
> 
>     2. The body has a line for some odd reason and the padding also seems off.
> 
>     3. The width of the page and the cards in the page feels strange.

How wide should the cards be?

garrett:
> How wide should the cards be?

Not sure.

There's a lot of blank space at the side, but the actions would be too far away. It's fine as is for now. I've posted a redesign of the page in #176 which will probably work around this issue by using expandable lists.

Issue #176, opened 2021-03-24
Change repository modal is odd


garrett:
The modal currently looks like this:

![Screenshot_2021-03-24 Software Updates - garrett Rain(3)](https://user-images.githubusercontent.com/10246/112339070-a03e1f00-8cbf-11eb-91f5-b20a7b7fe5eb.png)

![Screenshot_2021-03-24 Software Updates - garrett Rain(4)](https://user-images.githubusercontent.com/10246/112340577-f5c6fb80-8cc0-11eb-91f9-6004407d6219.png)

Instead, it should probably be using:

- data list: https://www.patternfly.org/v4/components/data-list
- inline editing: https://www.patternfly.org/v4/components/inline-edit

Or a table with editable rows: https://www.patternfly.org/v4/components/table#editable-rows

Since there's a lot of interaction inside the dialog with different states, we might even want to reconsider using a sub-page instead of a modal.

garrett:
Problems with the dialog:

1. Add new repository is a link, but should be a button. However, then it would clash with the other buttons in the dialog.
2. The inline editing is very weird and adds too much functionality in this dialog.
3. Adding a new repository closes this dialog and brings forth another... and dialogs should never ever call another one.
4. There's selection in the lists in the new PR, but this has nothing to do with selection... it has to do with which is active, not selected... selection is a temporary state to do something with an item, whereas active is a semi-permanent state.

However, redesigning this means we need to redesign the page itself so that it fits in properly too.

Here's a first start at that:

![cockpit-ostree-2021-03-29(1)](https://user-images.githubusercontent.com/10246/112834754-a6a21180-9098-11eb-8a4b-47e1462d1cfd.png)

I'm not very happy with the repository part yet, but this is better than what we have. There's probably a better set of widgets for it. Perhaps we really want to make it show only the current repository and make the editing behind some sort of disclosure widget or inline editing. I don't know. But the goal is to split apart the current repo dialog, so editing and adding can be their own separate dialogs, while the main part is not in a modal.

We could probably drop architecture from being visible and filter the branch to be the only arch (or arches if there is more than one) avaialble. But we don't have to worry about the i386/x86_64 split. Would we need to worry about an ARM HFP / AArch64 split though?

As for the … menus, the repository actions would be:

- Edit, Set active, Delete

And the deployment actions would be:

- Pin, Deploy, Delete
- Unpin, Deploy, Delete

(In addition to the main action of update + reboot / roll back + reboot.)

[...]

PR #178, opened 2021-03-25
UI related fixes here and there, towards using more PF4

KKoukiou:


martinpitt:
Code wise this LGTM. But as I honestly can't see the "off" spacing in #177 and what exactly "odd" is in #176, I'd rather let garrett do the official ack or nack. Thanks!

garrett:
> But as I honestly can't see the "off" spacing in #177 and what exactly "odd" is in #176, I'd rather let garrett do the official ack or nack. Thanks!

They're not consistent with the rest of Cockpit they're not standard with PatternFly, and they're not using generally accepted spacing rules with respect to design most everywhere.

Generally, when it comes to text, you want a certain amount of space above it and at the sides. Usually, there's more at the sides than the top. (GitHub and pretty much everything else does this too. Look at the buttons, for example.) It's because we read text horizontally and have spaces between words.

And you want a certain amount of space between icons and text. It's usually around 0.5rem / 1rem. But it should be consistent. The picture is obviously not consistent (and there's wayyyyyy too much space between the (!) and the text it is supposed to direct attention to).

The problem here is that it's an icon and then a link spaced as if they were separate items, instead of the same item. The icon + text in the second row is spaced together as if they belong together (and they do). Perhaps the top one has an icon and then a link instead of an icon in a link.

overview card

This one is not using the correct widgets or patterns we use elsewhere. There's a link with an icon that does an action. That should clearly be a button. The list has weird padding and spacing and doesn't look like other lists. We don't do editing like this anywhere else in Cockpit (or in PatternFly).

weird modal

A lot of the issues would hopefully be solved if we used the correct patterns and widgets, and used them consistently.

---

Anyway, I'm checking out the branch now to take a look at the changes. :wink:

garrett:
1. Add new repository is a link, but should be a button. However, then it would clash with the other buttons in the dialog.
2. The inline editing is very weird and adds too much functionality in this dialog.
3. Adding a new repository closes this dialog and brings forth another... and dialogs should never ever call another one.
4. There's selection in the lists in the new PR, but this has nothing to do with selection... it has to do with which is active, not selected... selection is a temporary state to do something with an item, whereas active is a semi-permanent state.

I think we need a redesign if we want to properly fix this. There's just too much going on here for a dialog; we probably need to make most of this a sub-page or a page section instead.

However, redesigning this means we need to redesign the page itself so that it fits in properly too.

Here's a first start at that:

cockpit-ostree-2021-03-29(1)

I'm not very happy with the repository part yet, but this is better than what we have. There's probably a better set of widgets for it. Perhaps we really want to make it show only the current repository and make the editing behind some sort of disclosure widget or inline editing. I don't know. But the goal is to split apart the current repo dialog, so editing and adding can be their own separate dialogs, while the main part is not in a modal.

We could probably drop architecture from being visible and filter the branch to be the only arch (or arches if there is more than one) avaialble. But we don't have to worry about the i386/x86_64 split. Would we need to worry about an ARM HFP / AArch64 split though?

As for the … menus, the repository actions would be:

- Edit, Set active, Delete

And the deployment actions would be:

- Pin, Deploy, Delete
- Unpin, Deploy, Delete

(In addition to the main action of update + reboot / roll back + reboot.)

We currently do not have pinning or unpinning capability in cockpit-ostree. This is how I think it would work, but it can be ignored and implemented later.

garrett (changes requested):
As far as this PR is concerned:

1. :+1:  The padding and spacing of the list has been much improved
2. :+1: I'm happy the `<hr>` line is gone
3. :-1: I'm not happy about using the table selection to indicate something is active... this was intended for selections to _do something with_

If you want to merge most of these changes in as-is without any big overhauls, I would be happy if we could rescope this PR to 1 & 2.

I think the dialog is a lost cause at this moment and we'd need to overhaul it to be like my mockup in the previous comment. But that might take time. So another PR would make sense for the larger changes.

garrett:
I've copied the bulk of my redesign comment over on issue #176.

KKoukiou:
@garrett agreed here - I removed the commits related to the dialog change

garrett:
The text in the heading doesn't line up to anything. It should either line up with the card's border, or the text of the card. (It's debatable which, but it doesn't line up with either now.)

Screenshot from 2021-04-08 18-41-31

The text of the card is inconsistent. Should the tabs line up with the heading and body text or should the other text line up with the tabs? (Same question, just which is anchored and which should be adjusted.)

Screenshot from 2021-04-08 18-42-03

Spacing of the headings is too large overall. But the available one where rolling back is possible is extremely odd and misaligned.

Screenshot_2021-04-08 Software Updates - garrett Rain(1)

Additionally, there's "Title Capitalization" used all over the place and we're supposed to use "Sentence caps" in PF4, but this possibly belongs to another PR.

garrett:
We'll want to redo the page for https://github.com/cockpit-project/cockpit-ostree/issues/176, but it may take a bit longer, so we could still fix the obvious brokenness above meanwhile.

KKoukiou:
Hey @garrett I think you forgot about this :) It contains some fixes, can you please take a look when you have time.

KKoukiou:
> The text in the heading doesn't line up to anything. It should either line up with the card's border, or the text of the card. (It's debatable which, but it doesn't line up with either now.)

Fixed, it was because of the toolbar padding.

> 
> The text of the card is inconsistent. Should the tabs line up with the heading and body text or should the other text line up with the tabs? (Same question, just which is anchored and which should be adjusted.)

Fixed, adding some margin for it. 


> 
> Spacing of the headings is too large overall. But the available one where rolling back is possible is extremely odd and misaligned.
>

Let's not address it in this commit. The paddings used are exactly the same as the documentation's https://www.patternfly.org/v4/components/data-list#actions-single-and-multiple

If you think these are wrong, as always let's open an issue upstream. 

 
> Additionally, there's "Title Capitalization" used all over the place and we're supposed to use "Sentence caps" in PF4, but this possibly belongs to another PR.

Right, another pR :)

garrett:
I'm getting errors when loading the page. Oops appears and the console has a LOT of spew.

I think it's because remotesList is null, yet attempts `includes()` in `changeRemoteModal.jsx`:
```jsx
<Button key="change-repo"
                variant="primary"
                isDisabled={!!editRepoDialogOpen || !remotesList.includes(selectedRemote)}
                onClick={() => {
            onChangeRemoteOrigin(selectedRemote).then(() => setIsModalOpen(false), ex => setError(ex.message));
        }}>
```

Here's the log (sorry; it's a screenshot... not really a good way to copy/paste it all):
Screenshot from 2021-04-12 14-55-00

KKoukiou:
@garrett unrelated to this PR the errors you see above.

garrett:
> @garrett unrelated to this PR the errors you see above.

Really? I cannot run this PR due to those errors. Are they in master too then?

garrett:
After switching to master, building, and seeing it work, I switched back to this branch and built it again and it worked. It's really odd, as I did a make clean after the first time too. :shrug:

garrett (approved):
>>    Spacing of the headings is too large overall. But the available one where rolling back is possible is extremely odd and misaligned.
>
> Let's not address it in this commit. The paddings used are exactly the same as the documentation's https://www.patternfly.org/v4/components/data-list#actions-single-and-multiple

But that's not the widget that should be used here? That's supposed to be a card title, not a list, right? :thinking: 

Everything _other than_ the heading is correct now. There are enough huge improvements that we should just get this in and work on a follow-up to properly fix the heading.

---

For reference, the heading still has the alignment and spacing issues:

Screenshot from 2021-04-12 15-53-35

- Orange: Space above vs. space below the button. It's the same. That's vertically centered.
- Lighter red: offset from baseline: 6px
- Darker red: difference between spacing above and below the text: 5px

---

But everything else (aside from the heading, mentioned above, and the dialog, mentioned in the linked issue) is good now. Thanks!

Approving for the fixes this is addressing.

PR #181, opened 2021-04-08
Cleanup unnecessary rhel cases

marusak:
Both rhel-atomic and continuous atomic are not a thing anymore.

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -41,25 +41,10 @@ CHECKOUT_LOCATION = "/var/local-tree"
 RPM_LOCATION = "/usr/share/rpm"
 KEY_ID = "95A8BA1754D0E95E2B3A98A7EE15015654780CBD"
 
-# old cockpit/ws container on rhel-atomic still has "[ ] Reuse my password for privileged tasks"
-login_opts = {}
-if os.environ.get("TEST_OS") == "rhel-atomic":
-    login_opts["legacy_authorized"] = True
-
 
 def get_list_item(index):
     # there are more <li> inside the item tables, so only count direct children
     return ".available-deployments > li:nth-child({0})".format(index)
-
-def get_os_name(m):
-    if m.image == "rhel-atomic":
-        return "rhel-atomic-host"
-    elif m.image == "fedora-coreos":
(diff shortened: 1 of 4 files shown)


test/check-ostree
```diff
@@ -41,25 +41,15 @@ CHECKOUT_LOCATION = "/var/local-tree"
...
-    if m.image == "rhel-atomic":
-        return "rhel-atomic-host"
-    elif m.image == "fedora-coreos":
+    if m.image == "fedora-coreos":
         return "fedora-coreos"
-    elif 'continuous' in m.image:
-        return 'centos-atomic-host'
-    else:
-        return "fedora-atomic"
+    return "fedora-atomic"
```
martinpitt, line 52: Fedora Atomic also hasn't been a thing for a while. Want to clean that up as well while you are at it? I think this should rather `raise NotImplementedError` here.

martinpitt (commented):
Cheers!

test/check-ostree
marusak: I was wondering what should be the fallback. Is there any chance that we will introduce anything else? If not, then I can just replace calls to this function with directly adding `fedora-coreos` into the code.

martinpitt: Right now there's only fedora-coreos. RHEL coreos isn't a standalone supported product. We may add this at some point, but as there is nothing planned right now, I'd just make it as simple as possible, and we can reintroduce an indirection if/when we need it.

test/check-ostree
```diff
@@ -247,7 +230,7 @@ class OstreeRestartCase(MachineCase):
 
         # Check update data
         b.wait_visible("{0} div.tree".format(get_list_item(1)))
-        b.wait_in_text("{0} .os".format(get_list_item(1)), name)
+        b.wait_in_text("{0} .os".format(get_list_item(1)), "fedora-coreos")
```
martinpitt, line 233: ah, I didn't think of that usage here -- that repeats the OS name a lot here. But *shrug*, we can re-factorize it later on as well.

martinpitt (approved):
Thank you!

PR #186, opened 2021-06-23
webpack: bring back sass-loader, using Dart Sass

KKoukiou:
Cherry-picked cockpit commit 012483503e53ac069a44c569374ea46756ecb850

pkg/lib was bumped as well.


package.json
```diff
@@ -41,6 +41,8 @@
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "po2json": "^1.0.0-alpha",
     "qunit": "^2.9.3",
+    "sass": "~1.32.0",
```
allisonkarlitskaya, line 44: Same comment as ostree: we might be able to get away with unpinning this with new PF4.

martinpitt: ostree is pure PF4, and PR #187 just updated to latest PF4, so that might indeed work now. In the current PR, mini-css-extract-plugin barfs, so this needs more work anyway.

package.json
```diff
@@ -38,15 +39,16 @@
     "htmlparser": "^1.7.7",
     "jed": "^1.1.1",
     "mini-css-extract-plugin": "^0.11.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.3",
     "po2json": "^1.0.0-alpha",
```
allisonkarlitskaya, line 42: you can probably replace that with `gettext-parser`: the cockpit po plugin doesn't use po2json anymore

allisonkarlitskaya (changes requested):
One minor tweak.  Forgot to mention that on the other things that got bumped to the new `pkg/lib` unfortunately :(

package.json
allisonkarlitskaya: skobyda/cockpit-certificates#58 and cockpit-project/cockpit-podman#731

allisonkarlitskaya (approved):
I'm not the best person to review this, but I have no further objections :)

martinpitt (approved):
I did a visual test, and it looks exactly as before. Thanks for the nice "all around" update!

PR #189, opened 2021-07-05
various: master → main updates

allisonkarlitskaya:
Also update the weblate repo reference to `main`


allisonkarlitskaya:
Nobody here has credentials to change the branch used by the ostree project on weblate, so let's keep the weblate branch with the `master` name for now and drop the commit here which changed that. When @marusak gets back we can finish this up.

martinpitt:
Looks ok, but no tests are triggered -- blocked on a testmap update.

martinpitt (approved):

PR #192, opened 2021-07-15
Port from moment to date-fns

martinpitt:
moment is deprecated now. Use cockpit's new timeformat library instead,
which wraps date-fns and deals with localization automatically.

This fixes non-localized date/time formatting, as the page forgot to
call moment.locale().

----

There is no visual difference in English:

main

It just looks different in other languages now due to the previously missing localization.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,14 +164,14 @@ bots:
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
 	flock Makefile sh -ec '\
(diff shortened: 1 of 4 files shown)


martinpitt:
Adjusted the test to new testlib ([failure log](https://logs.cockpit-project.org/logs/pull-192-20210715-050650-deb9e6df-fedora-coreos/log.html#2-2))

martinpitt:
@KKoukiou : Friendly review reminder :hugs:

KKoukiou (approved):

PR #200, opened 2021-09-24
package.json: Add new dependencies

marusak:
For some reason I could not build this without adding these deps.
If someone can explain to me why this works in CI but not on my machine I'll be glad :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,9 +54,11 @@
     "@patternfly/patternfly": "4.135.2",
     "@patternfly/react-core": "4.157.3",
     "@patternfly/react-styles": "4.11.16",
+    "@patternfly/react-icons": "4.11.17",
     "date-fns": "2.24.0",
     "react": "16.14.0",
     "react-dom": "16.14.0",
+    "react-dropzone": "11.4.2",
     "throttle-debounce": "2.3.0"
   }
 }


package.json
```diff
@@ -54,9 +54,11 @@
     "@patternfly/patternfly": "4.135.2",
     "@patternfly/react-core": "4.157.3",
     "@patternfly/react-styles": "4.11.16",
+    "@patternfly/react-icons": "4.11.17",
```
martinpitt, line 57: That is fully correct -- it's a dependency anyway, and we really ought to track the version explicitly. This has bitten us in cockpit and other projects already, as not all versions are compatible with each other.

package.json
```diff
@@ -54,9 +54,11 @@
...
     "date-fns": "2.24.0",
     "react": "16.14.0",
     "react-dom": "16.14.0",
+    "react-dropzone": "11.4.2",
```
martinpitt, line 61: This I don't understand, and agree with -- nothing in our own code uses react-dropzone, so we shouldn't directly depend on it. Looking at my
[package-lock.json](https://github.com/cockpit-project/cockpit-ostree/files/7226063/package-lock.json.txt), I see that it is a dependency of `@patternfly/react-core`, and it wants version 9.0.0, not 11.4. So this seems actively wrong to me?

Can you compare your package-lock.json (without this PR) to mine, and check if there is some meaningful difference?

I'm running in cockpit/tasks container, with npm-6.14.15-1.14.17.6.1.fc34.x86_64 and nodejs-14.17.6-1.fc34.x86_64

martinpitt (changes requested):

martinpitt:
How does it actually fail for you?

package.json
marusak: If I remove it, it fails like this:
```
ERROR in ./node_modules/@patternfly/react-core/dist/esm/components/FileUpload/FileUpload.js 3:0-38
Module not found: Error: Can't resolve 'react-dropzone' in '/home/mmarusak/cockpit/cockpit-ostree/node_modules/@patternfly/react-core/dist/esm/components/FileUpload'
```
npm version `7.21.1`
node version `16.9.1`

martinpitt: That's Fedora 35 ? Does it actually install react-dropzone, and how does it look like in your package-lock.json?

martinpitt: Does that affect c-{podman,machines} and cockpit itself also? We don't use the `FileUpload` component in any of our projects, so ATM I don't see what's special about ostree

marusak: It does not affect any other component. I am not sure, why react-icons need to be specified explicitly either. I was trying to understand what's the difference between ostree and podman, but could not find anything noteworthy.

martinpitt: We do use react-icons explicitly:
```
src/app.jsx:import { ExclamationCircleIcon, OutlinedCheckCircleIcon } from '@patternfly/react-icons';
src/changeRemoteModal.jsx:import { PencilAltIcon, AddCircleOIcon } from '@patternfly/react-icons';
```
Thus we should have an explicitly versioned `dependencies` entry for it. That's not the root cause for the build failure (@jelly had some good ideas about our webpack config's `module.export.resolve` entry?), but it's still the right thing to do.

marusak: ha, now I changed branch in podman (to something new I presume) and it fails as well:
```
ERROR in ./node_modules/@patternfly/react-core/dist/esm/components/FileUpload/FileUpload.js 3:0-38
Module not found: Error: Can't resolve 'react-dropzone' in '/home/mmarusak/cockpit/cockpit-podman/node_modules/@patternfly/react-core/dist/esm/components/FileUpload'

```

jelly:
> How does it actually fail for you?

I've seen the same failure here, which makes me believe webpack can't import nested node_modules:

```
[jelle@natrium][~]%npm -v
7.24.1
[jelle@natrium][~]%node -v
v16.10.0
```

```
[jelle@natrium][~/projects/cockpit-ostree]%make
NODE_ENV= npm run build

> build
> webpack

ERROR in ./node_modules/@patternfly/react-core/dist/esm/components/FileUpload/FileUpload.js 3:0-38
Module not found: Error: Can't resolve 'react-dropzone' in '/home/jelle/projects/cockpit-ostree/node_modules/@patternfly/react-core/dist/esm/components/FileUpload'

webpack compiled with 1 error
make: *** [Makefile:76: dist/manifest.json] Error 1
[jelle@natrium][~/projects/cockpit-ostree]%npm ls react-dropzone
cockpit-ostree@ /home/jelle/projects/cockpit-ostree
└─┬ @patternfly/react-core@4.157.3
  └── react-dropzone@9.0.0

[jelle@natrium][~/projects/cockpit-ostree]%fd --no-ignore-vcs react-dropzone
node_modules/@patternfly/react-core/node_modules/react-dropzone
node_modules/@patternfly/react-core/node_modules/react-dropzone/typings/react-dropzone.d.ts
[jelle@natrium][~/projects/cockpit-ostree]%vim webpack.config.js
[jelle@natrium][~/projects/cockpit-ostree]%cp -r node_modules/@patternfly/react-core/node_modules/react-dropzone node_modules
[jelle@natrium][~/projects/cockpit-ostree]%make
NODE_ENV= npm run build

> build
> webpack

webpack compiled successfully
```

PR #207, opened 2021-11-01
Prevent [Object object] from being shown

jelly:
When switching from administrative to limited the exception is an array
which prints [Object object] as console.message prints an array.


martinpitt (approved):
Danke!

PR #213, opened 2021-11-19
React to superuser changes

martinpitt:
This is just very coarse for now, but better than a broken page.

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -622,6 +622,10 @@ class OstreeCase(MachineCase):
         b.wait_in_text(".pf-c-empty-state__body", "Not authorized")
         self.assertIn("Reconnect", b.text(".pf-c-empty-state button"))
 
+        b.become_superuser()
+        b.switch_to_frame("cockpit1:localhost/updates")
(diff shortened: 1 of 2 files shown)


test/check-ostree
```diff
@@ -611,30 +611,19 @@ class OstreeCase(MachineCase):
...
-
-        b.enter_page("/updates")
-
+        self.login_and_go("/updates", superuser=False)
         b.wait_in_text(".pf-c-empty-state__body", "Not authorized")
         self.assertIn("Reconnect", b.text(".pf-c-empty-state button"))
-        self.allow_journal_messages("cannot reauthorize identity.*")
+
+        b.become_superuser()
+        b.wait_visible('.available-deployments')
```
marusak, line 626: tests don't seem to be happy with this

martinpitt: Ugh, worked locally.. the screenshot looks like some `expect_load()` ought to happen, but I tried that and it doesn't work. I'll investigate.

marusak (approved):
Thanks!

PR #215, opened 2021-11-19
Makefile: Fix name of node cache tarball

martinpitt:
"cockpit-" prefix was duplicated.


martinpitt:
This [broke the release](https://github.com/cockpit-project/cockpit-ostree/actions/runs/1480526450), as the asciibetically first tarball name was now the cockpit-cockpit-ostree-node-cache* one, which does not contain the spec file.

KKoukiou (approved):

PR #246, opened 2022-04-20
test: Add pixel tests

martinpitt:
Cover the toolbar, a deployment, and the remote management dialog.

---

We get a lot of PF updates now (since switching to pre-release), and validating them visually is too much of a chore now. This coverrs the three main UI items, so we should be ok with landing them without visual inspection.


martinpitt:
Resolved conflict/rebased after PR #244. I wanted to do this anyway, to make sure this works with latest test API.

martinpitt:
Aha, [failed](https://logs.cockpit-project.org/logs/pull-246-20220420-073015-9c734a08-fedora-coreos/log.html) on the missing medium/mobile pixels. Added.

PR #247, opened 2022-04-22
test: switch to new container image

jelly:
The bitnami image mirrored from dockerhub was removed from quay.io and
as we can't pull from dockerhub due to rate limitting use a different
nginx image. In the future the image should be pulled into the testing
container so tests can run fully offline.

Note, I've attempted to use the smaller lighttpd image but this image does not work well with our tests, the container isn't getting killed with `stop_trivial_httpd`. No Python image seems to be available on quay.io for us to simply use python -m http.server


martinpitt (approved):
Cheers! https://quay.io/repository/jitesoft/nginx?tab=info looks legit, and it is updated reasonably often.

PR #248, opened 2022-04-22
test: autopep8 check-ostree

jelly:


martinpitt (approved):
Thanks! We may consider adopting https://github.com/cockpit-project/cockpit-machines/pull/671 and https://github.com/cockpit-project/cockpit-machines/pull/679 to run test/static-code on this project.

PR #249, opened 2022-04-22
test: offline image-customize

jelly:
Create the test virtual machine offline now that the coreos image
contains the used nignx container. For rpm-ostree to run offline, invoke
it with --cache-only so it doesn't try to fetch repository data.


martinpitt:
Image refresh landed, retrying "normal" test.

martinpitt (approved):

PR #263, opened 2022-05-31
Release packit/workflow yadayada

martinpitt:
This one is not 100% identical to podman. We don't need some hacks (as it doesn't have gating tests), and I needed an extra spec fix.

[test release](https://github.com/martinpitt/cockpit-ostree/releases/tag/190.2), [workflows](https://github.com/martinpitt/cockpit-ostree/runs/6674798841), [dist-git PR](https://src.fedoraproject.org/rpms/cockpit-ostree/pull-request/4)


allisonkarlitskaya (approved):
ya ya ya

PR #281, opened 2022-09-14
Install into /usr/local/share/ by default

martinpitt:
/usr is package manager territory, and not even writable on OSTree based systems such as Fedora CoreOS or RHEL Edge. Given that we primarily target OSTree, `make install` should do the right thing there.

You can still override this with `make install PREFIX=...`. Use that in the RPM spec file.

---

Motivated by https://bugzilla.redhat.com/show_bug.cgi?id=2126471


marusak (approved):
Thanks :)

PR #285, opened 2022-10-12
.github: update checkout action to v3

jelly:
actions/checkout@v3 uses nodejs 16 as runtime instead of the deprecated nodejs 12 runtime.


martinpitt:
And no peer review? :grin:  But retroactively +1, thank you!

PR #291, opened 2022-11-04
Dark theme update

jelly:


.eslintrc.json
```diff
@@ -39,6 +39,8 @@
 
         "no-console": "off",
         "quotes": "off",
+	"key-spacing": "off",
+	"object-property-newline": "off",
```
martinpitt, line 43: Hmm, IMHO it would be better to entirely ignore pkg/lib/ (just like node_modules) in eslint for all cockpit extensions. But that's not a blocker.

martinpitt (approved):
:dark_sunglasses:

.eslintrc.json
jelly: Oh, good point! Meanwhile this is underway https://github.com/cockpit-project/cockpit/pull/17876 Let's make it consistent everywhere!

jelly: Well, that was easier then expected and we do this already in cockpit-podman...

martinpitt (approved):
Yay, thanks!

PR #323, opened 2023-02-07
test with python bridge

martinpitt:
Build the Python bridge from the cockpit project and install it into the VM so we can run tests against it.

The fedora-coreos image does not have pip, and rpm-ostree install makes it unwieldy to install on the fly as it has so many dependencies (which can't be updated if they are installed in an older version).

So instead of building and installing a wheel, copy our two modules into the site directory, and create a /usr/local/bin/cockpit-bridge wrapper. This is a giant hack, but only temporary until we package the pybridge into cockpit-bridge.rpm.

---

 - [x] https://github.com/cockpit-project/cockpit/pull/18296
 - [x] https://github.com/cockpit-project/cockpit/pull/18314
 - [x] Fix (https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3272) ; got into F37 and F38 now
 - [x] https://github.com/cockpit-project/bots/issues/4669 to pick it up in our fedora-coreos image
 - [ ] RHEL/CentOS 9 backport: https://bugzilla.redhat.com/show_bug.cgi?id=2217771

I know @allisonkarlitskaya is allergic to this, and I am allergic to installing pip (or any other build-time dependency) directly onto our image. I don't want to spend time with a temporary spec which builds cockpit-bridge.rpm with the pybridge in c-ostree's Makefile, so I propose to just keep this around as a draft and test c-ostree occasionally here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,12 +139,36 @@ rpm: $(TARFILE) $(NODE_CACHE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
+# pybridge scenario: build and install the python bridge from cockpit repo
+# this is a giant hack to test the pybridge in a scenario before it gets packaged into cockpit-bridge.rpm
+ifeq ("$(TEST_SCENARIO)","pybridge")
+COCKPIT_PYBRIDGE_REF = main
+
+tmp/pybridge:
+	# no pip on FCOS
+	rm -rf tmp/pybridge
+	git init tmp/pybridge
+	git -C tmp/pybridge remote add origin https://github.com/cockpit-project/cockpit
+	git -C tmp/pybridge fetch --depth=1 origin ${COCKPIT_PYBRIDGE_REF}
+	git -C tmp/pybridge reset --hard FETCH_HEAD
+	tmp/pybridge/modules/checkout
+
+VM_DEPENDS = tmp/pybridge
+VM_CUSTOMIZE_FLAGS = --upload tmp/pybridge/modules/systemd_ctypes/src/systemd_ctypes:/tmp/pymods \
+	--upload tmp/pybridge/modules/ferny/src/ferny:/tmp/pymods \
+	--upload tmp/pybridge/src/cockpit:/tmp/pymods \
+	--run-command 'P=$$(python3 -c "import sysconfig; print(sysconfig.get_path('"'"'purelib'"'"'))"); \
+	               mkdir -p $$P; cp -r /tmp/pymods/* $$P' \
+	--run-command 'echo "python3 -m cockpit.bridge \"\$$@\"" > /usr/local/bin/cockpit-bridge; chmod a+x /usr/local/bin/cockpit-bridge' \
+	--upload cockpit-askpass:/usr/local/bin/cockpit-askpass
+endif
+
 # build a VM with locally built rpm installed, cockpit/ws container, and local
 # ostree for testing
-$(VM_IMAGE): rpm bots
+$(VM_IMAGE): rpm bots $(VM_DEPENDS)
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
 	bots/image-customize -v --upload $$(ls $(PACKAGE_NAME)-*.noarch.rpm):/tmp/ \
-		--no-network \
+		--no-network $(VM_CUSTOMIZE_FLAGS) \
 		--run-command 'rpm -q cockpit-ostree && rpm-ostree --cache-only override replace /tmp/*.rpm || rpm-ostree --cache-only install /tmp/*.rpm' \
 		$(TEST_OS)
 	# building the local tree needs the modified tree from above booted already
--- a/cockpit-askpass
+++ b/cockpit-askpass
@@ -1,0 +1,4 @@
+#!/usr/bin/python3
+import sys
+from cockpit._vendor.ferny.interaction_client import main
+sys.exit(main())


martinpitt:
Aside from the "unsupported browser: pybridge" copy-pasta bug (which we already fixed everywhere else), this mostly fails on `sudo: unable to run /nonexistent/libexec/cockpit-askpass`. I.e. the pybridge can't detect libexecdir properly on coreos.

This is reflected here:
```
# cockpit-bridge --packages
cockpit.packages-WARNING: Could not detect libexecdir
base1                                                         /usr/share/cockpit/base1
[...]
```

That's because we check for `cockpit-certificate-helper`, but that's packaged in cockpit-ws only (which we don't install). I fixed that in https://github.com/cockpit-project/cockpit/pull/18296 and will add a temporary commit here to pull in that SHA instead. Locally this gets me some passing tests.

Makefile
```diff
@@ -104,12 +104,34 @@ rpm: $(RPMFILE)
...
+	git -C tmp/pybridge fetch --depth=1 origin ${COCKPIT_PYBRIDGE_REF}
+	git -C tmp/pybridge reset --hard FETCH_HEAD
+	tmp/pybridge/tools/systemd_ctypes
+
+VM_DEPENDS = tmp/pybridge
+VM_CUSTOMIZE_FLAGS = --upload tmp/pybridge/modules/systemd_ctypes/src/systemd_ctypes:/tmp/pymods \
+	--upload tmp/pybridge/src/cockpit:/tmp/pymods \
+	--run-command 'P=$$(python3 -c "import sysconfig; print(sysconfig.get_path('"'"'purelib'"'"'))"); \
+	               mkdir -p $$P; cp -r /tmp/pymods/* $$P' \
+	--run-command 'echo "python3 -m cockpit.bridge \"\$$@\"" > /usr/local/bin/cockpit-bridge; chmod a+x /usr/local/bin/cockpit-bridge'
```
allisonkarlitskaya, line 128: Please use either pip or python3-installer for this.

allisonkarlitskaya: Specifically: we could modify the install command for wheels to check if /usr is writable and install to /usr/local otherwise.  Or just always /usr/local.

martinpitt: I already had a local change to `image-customize` to install into /usr/local/ -- fully agreed that we wan this. This would be conceptually the right thing to do, plus it would allow us to more easily switch between C and py bridge in a prepared image (something that I often sorely miss). However, a lot of our tests hardcode/assume /usr/bin/cockpit-bridge, so we'll need some fixes first.

Neither pip nor -installer are part of fedora-coreos, and these will not be runtime dependencies of cockpit-bridge, so I don't want to do this. That's what I described in the description -- but that's fine for now. We can agree to disagree and leave this as a draft PR. The utility is to have a vehicle to run and investigate tests, not so much in actually landing this PR.

allisonkarlitskaya: https://github.com/cockpit-project/cockpit/pull/18301

martinpitt: FTR: I find that waaaaaay to much overhead, and carrying around untested/half-tested dead weight in cockpit.git. I only want a draft PR as a vehicle to investigate the ostree failures, this is good enough. (Again: I don't need/want to land this PR)

martinpitt:
The remaining failures are all because the pybridge does not support connecting to named (custom) D-Bus addresses. rpm-ostree unfortunately requires that. @mvollmer and @allisonkarlitskaya are working on this, so blocking further investigations on that.

allisonkarlitskaya:
I also intend to keep blocking this ;)

If we want this as a proof of concept or something to play with sure, but the Makefile changes here are really a disaster.  I don't want to approve them.

I think a way forward, but it's a bit more work: I would conceptually support modifying `image-customize` to teach it about how to install wheels on a host which lacks `pip` and `python3-installer`.  These things are basically zipfiles that need to be unpacked into a special location, plus the secret sauce for script creation.  It's not *so* complicated.

We could also use pip or python3-installer on the host (both of which will soon be in the tasks container) along with a temporary directory and then rsync the result over to the VM.  We'd just need to figure out a way to hack the directory for the correct python version in that case.

Or we could just leave this PR as it is.

martinpitt:
https://github.com/cockpit-project/cockpit/pull/18314 changed the failure modes, but still all but one tests fail. It seems connecting to the bus address now succeeds, and it starts the transaction, but it never gets the result:

```
dbus: {"call":["/org/projectatomic/rpmostree1/Sysroot","org.projectatomic.rpmostree1.Sysroot","ReloadConfig",[]],"id":"8"} ostree.js:2:331959
dbus: {
  "reply": [
    []
  ],
  "id": "8",
  "flags": null,
  "type": null
}
ostree.js:2:331959
dbus: {"call":["/org/projectatomic/rpmostree1/fedora_coreos","org.projectatomic.rpmostree1.OS","DownloadUpdateRpmDiff",[]],"id":"9"} ostree.js:2:331959
dbus: {
  "reply": [
    [
      "unix:abstract=/tmp/rpm-ostree/dbus-rBOvkVIk"
    ]
  ],
  "id": "9",
  "flags": null,
  "type": null
}
ostree.js:2:331959
dbus open:  
Object { superuser: true, address: "unix:abstract=/tmp/rpm-ostree/dbus-rBOvkVIk", bus: "none", payload: "dbus-json3" }
ostree.js:2:331959
dbus: {"add-match":{"path":"/"}} ostree.js:2:331959
dbus: {"call":["/","org.projectatomic.rpmostree1.Transaction","Start",[]],"id":"2"} ostree.js:2:331959
dbus: {
  "notify": {
    "/org/projectatomic/rpmostree1/Sysroot": {
      "org.projectatomic.rpmostree1.Sysroot": {
        "ActiveTransactionPath": "unix:abstract=/tmp/rpm-ostree/dbus-rBOvkVIk",
        "ActiveTransaction": [
          "DownloadUpdateRpmDiff",
          ":1.30",
          "/org/projectatomic/rpmostree1/fedora_coreos"
        ]
      }
    }
  }
}
2 ostree.js:2:331959
dbus ready: 
Object { superuser: true, address: "unix:abstract=/tmp/rpm-ostree/dbus-rBOvkVIk", bus: "none" }
ostree.js:2:331959
dbus: {
  "owner": null
}
cockpit.js:1:27043

```
No answer, ever. Could it be that the "call" is done before dbus says "ready"?

I pushed another hack commit to disable preloading, as this is just unnerving with `window.debugging = "dbus"`. I am running

     rm -f Ostree*; TEST_OS=fedora-coreos TEST_BROWSER=firefox TEST_SHOW_BROWSER=1 TEST_SCENARIO=pybridge test/check-ostree OstreeCase.testRemoteManagement -stv

Once the VM is running, you have to `systemctl stop rpm-ostreed` between each attempt to cancel the currently running transaction, then click "check for updates".

martinpitt:
I have a feeling that this is https://github.com/systemd/systemd/pull/26375 all over again. [rpm-ostree is gdbus](https://github.com/coreos/rpm-ostree/blob/main/src/daemon/rpmostreed-transaction.cxx#L36).

martinpitt:
I tested cockpit-ostree against https://github.com/cockpit-project/cockpit/pull/18314 and https://github.com/systemd/systemd/pull/26375 . Now all but one tests succeed, and the last one fails on an unrelated issue.

martinpitt:
I cherry-picked the glib branch onto Fedora 37's glib 2.74.1 version (applies cleanly), uploaded it to my [test COPR](https://copr.fedorainfracloud.org/coprs/martinpitt/test-fixes/build/5533842/)], `rpm-ostree override replace`'d it on our current Fedora CoreOS image, and ran the tests. They are a *lot* happier now! Only `OstreeRestartCase.testOstree` gets confused by the additional ostree commit due to the layered package, but that's expected -- and the other 4 succeed.

martinpitt:
Good news! A fixed glib is [finally in F37](https://bodhi.fedoraproject.org/updates/FEDORA-2023-1fcb3adca7), and F38 already got the fix a few weeks ago. So now the question is just whether CoreOS will pick up the fix via updating F37 or moving to F38 first :grin:

martinpitt:
Actually, [CoreOS](https://builds.coreos.fedoraproject.org/streams/testing.json) just switched over to F38 very recently, so let's refresh and test again!

martinpitt:
This is broken heavily right now, but pushing here to avoid losing my (non-trivial) rebase.

martinpitt:
A-ha! Was missing cockpit-askpass, so it couldn't get privs anywhere. Works fine now locally, let's see whether CI agrees.

martinpitt:
Yay! That's our last project, now they all work with the pybride :+1:

martinpitt:
https://github.com/cockpit-project/bots/pull/5074 slurped the Python bridge into our fedora-coreos image, so this is moot now. 

The [RHEL 9 backport](https://bugzilla.redhat.com/show_bug.cgi?id=2217771) also landed, so everything done :heavy_check_mark:

PR #325, opened 2023-02-08
workflows: Pacify git's permission check in release workflow

jelly:
Similar to https://github.com/cockpit-project/bots/pull/4224


martinpitt (approved):
:adhesive_bandage:

PR #330, opened 2023-02-14
manifest: Add condition

### LINKED ISSUE #295: Improve improper cockpit-ostree install on systems without rpm-ostree
We've been getting a lot of issues lately about people who have cockpit-ostree installed on a system that isn't based on rpm-ostree.

Whether someone installed `cockpit-*` or a weird dependency chain might've pulled in `ostree` (and then `cockpit-ostree`) on their system, we can improve the experience.

This can range anywhere from:
- improving the error message to be more clear about what the problem is and suggest a solution
- checking the system and switching to the standard UI if the system is not based on `rpm-ostree` with a message (perhaps even automatically with a [small banner](https://www.patternfly.org/v4/components/banner/design-guidelines) at the top with an info message, so there's some UI with the issue)

We might even want to do this in steps, like immediately improve the error messaging and then work on a follow-up to handle it more gracefully.

---

A few recent issues within the past few days, as an example:

- https://github.com/cockpit-project/cockpit/issues/17902
- https://github.com/cockpit-project/cockpit/discussions/17606
- https://github.com/cockpit-project/cockpit/issues/17858

martinpitt:
Only show ostree if the system is actually *using* OSTree, by checking for /sysroot/ostree. It sometimes happens that users install cockpit-ostree on classic RPM systems, which "breaks" the "Software Updates" page.

This is a better solution than the current static priorities of packagekit vs. ostree. However, this only works with the Python bridge [1].

Fixes #295

[1] https://github.com/cockpit-project/cockpit/pull/18291

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -3,6 +3,10 @@
     "requires": {
         "cockpit": "125"
     },
+    "conditions": [
+        {"path-exists": "/usr/bin/rpm-ostree"},
+        {"path-exists": "/sysroot/ostree"}
+    ],
 
     "tools": {
         "index": {


src/manifest.json
```diff
@@ -3,6 +3,10 @@
     "requires": {
         "cockpit": "125"
     },
+    "conditions": [
+        {"path-exists": "/usr/bin/rpm-ostree"},
```
martinpitt, line 7: We don't strictly need this I think, but it's cheap and communicates the intention.

jelly (approved):

src/manifest.json
jelly: Right, it's a dependency of the package already I assume, but sounds fair.

martinpitt: Yes, but the point of this is to make our pages work without packaging, i.e. using them on target hosts without cockpit-bridge and any installed packages.

garrett:
:tada: 

> this only works with the Python bridge

All the more reason to completely switch to the Python bridge soon. :wink:

humantex:
I'm unsure where in the commit process this might be, but just a quick update that cockpit-ostree is still showing in the update queue as part of the software update page. Forgive the rough formating, as this is a straight copy/paste from the expanded view of all affected packages. For reference: currently running Cockpit 284, on Fedora 37 Workstation prior to starting this update. Once again, removing `cockpit-ostree` afterwards puts everything right.

```
cockpit-composer	
44-1.fc37
Automatic update for cockpit-composer-44-1.fc37.
Packages
cockpit-composer
Automatic update for cockpit-composer-44-1.fc37.

Changelog for cockpit-composer
* Thu Feb 23 2023 Packit <hello@packit.dev> - 44-1
- Minor bug fixes
- Update translations

* Wed Feb 15 2023 Packit <hello@packit.dev> - 43-1
- Add customizations
- Fix bugs

* Mon Nov 28 2022 Packit <hello@packit.dev> - 42-1
- Update most components to use React 17 and PF4
- Update store to user redux-toolkit
- Update all style, build, and translations config
- Update all NPM dependencies
- Readd edge-simplified-installer image type

cockpit-machines	
284-1.fc37
Automatic update for cockpit-machines-284-1.fc37.
Packages
cockpit-machines
Automatic update for cockpit-machines-284-1.fc37.

Changelog for cockpit-machines
* Wed Feb 22 2023 Packit <hello@packit.dev> - 284-1
- Stability and performance improvements

cockpit-ostree	
1:192-1.fc37
Automatic update for cockpit-ostree-192-1.fc37.
Packages
cockpit-ostree
Automatic update for cockpit-ostree-192-1.fc37.

Changelog for cockpit-ostree
* Wed Feb 08 2023 Packit <hello@packit.dev> - 1:192-1
- Stability and performance improvements
```

martinpitt:
@humantex : We didn't change any dependencies. IMHO it is correct to recommend cockpit-ostree if rpm-ostree is installed. If you don't want that, then please remove rpm-ostree as well. What we did here is to *hide* the ostree version of "Software Updates" in cockpit if the system does not actually *use* OSTree (like yours). However, that change isn't effective yet, as it only works with the upcoming Python bridge, not the C bridge yet. Sorry!

humantex:
Thank you for the comment @martinpitt. I assumed it might still be in-process.

On removing `rpm-ostree`... It wasn't something I explicitly installed and don't recall if I ever saw anything to announce it as a dependency. Is there a way to find out if it really is a required package for whatever is currently installed? If I don't really need it, it's gone - as long as it doesn't cascade into other issues I have no problem in removing it.

martinpitt:
@humantex : You can try with `rpm -e --test rpm-ostree`. Also `sudo dnf remove` will ask you for confirmation anyway.

humantex:
Thank you @martinpitt : The 'test' line worked very well, and using it in several successive commands I could walk back through the tree to find the root and reason `rpm-ostree` was originally installed.

```
~]$ rpm -e --test rpm-ostree
error: Failed dependencies:
        rpm-ostree is needed by (installed) osbuild-ostree-78-1.fc37.noarch
        rpm-ostree(x86-64) is needed by (installed) gnome-software-rpm-ostree-43.5-1.fc37.x86_64

~]$ rpm -e --test osbuild-ostree
error: Failed dependencies:
        osbuild-ostree >= 70 is needed by (installed) osbuild-composer-worker-74-1.fc37.x86_64

~]$ rpm -e --test osbuild-composer-worker
error: Failed dependencies:
        osbuild-composer-worker = 74-1.fc37 is needed by (installed) osbuild-composer-74-1.fc37.x86_64

~]$ rpm -e --test osbuild-composer
error: Failed dependencies:
        osbuild-composer >= 28 is needed by (installed) cockpit-composer-44-1.fc37.noarch
```

Although I haven't completed the first attempt at it, I wanted to build something to mirror my own system for others to use as a Fedora-based DNS Trap 'virtual appliance', and use `cockpit-compose` as the frontend to build/maintain it. As long as Image Builder (`cockpit-composer`) requires it, I can't simply remove `rpm-ostree` and solve it for the long term. It's still simple enough for me to remove `cockpit-ostree` in the event it gets (re)installed, but I won't assume that approach will work for everyone.

I could raise this as a concern with the maintainer of `cockpit-composer`, but I'm unsure if that would lead to a resolution with Cockpit, as the dependencies that `osbuild-composer` requires may not be as easy to trim without farther reaching issues.

martinpitt:
@humantex : There's also a relatively simple workaround for you: Create /etc/cockpit/packagekit.override.json with `{ "priority": -1 }` should do.

PR #333, opened 2023-02-16
Add Github code scanning workflow

jelly:
This makes ostree 100% green \o/

### DIFF
--- a/.github/codeql-config.yml
+++ b/.github/codeql-config.yml
@@ -1,0 +1,2 @@
+paths:
+  - src
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,33 @@
+name: CodeQL
+on: [push, pull_request]
+
+jobs:
+  analyze:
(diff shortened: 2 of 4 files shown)


martinpitt:
LGTM, thanks! That [pixel failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-333-20230216-115104-bf813a3d-fedora-coreos/log.html) is very odd, though -- it can't possibly that duplicated `rel=`, can it? Retrying once for comparison, this could also just be a random flake -- but I've never seen the ostree tests flake like that.

jelly:
It moved again! This is really interesting!

martinpitt:
@jelly : But at first sight it seems stable between the two runs? Want to just update it and confirm that it's stable?

jelly:
> @jelly : But at first sight it seems stable between the two runs? Want to just update it and confirm that it's stable?

Soo... the pixel tests was not rebased! *SHOCK*

martinpitt (approved):
Nice!

PR #340, opened 2023-03-01
test: Wait until deployment table is stable

mvollmer:


mvollmer:
This is only a theory at this point... I am not sure if this really happens...

martinpitt (approved):
Good theory! Even if it wouldn't help the pixel tests, it's still a good thing to do.

PR #341, opened 2023-03-02
test: Ignore unstable deployment table columns

mvollmer:
The previous try to wait for the table to become stable didn't work, since the table appears fully populated in one step.  The unstable part that influences the column widths is the name of the second deployment, which changes from one run to the next.

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -183,13 +183,6 @@ class OstreeRestartCase(MachineCase):
         b.login_and_go("/updates")
         b.enter_page("/updates")
 
-        # Wait until all two deployments are shown.  The second one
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Ack -- completely ignoring the label feels a bit odd, but let's postpone that to the ongoing redesign -- that will perhaps have a more stable position then.

mvollmer:
> Ack -- completely ignoring the label feels a bit odd, but let's postpone that to the ongoing redesign -- that will perhaps have a more stable position then.

I still like this approach: https://github.com/cockpit-project/cockpit/pull/17654

Issue #300, opened 2022-12-06
support ostree native containers (container as OS)


garrett:
Details: https://coreos.github.io/rpm-ostree/container/

> rpm-ostree inherits work in [ostree-rs-ext](https://github.com/ostreedev/ostree-rs-ext/) to create “container native ostree” functionality. This elevates OCI/docker containers to be natively supported as a transport mechanism for bootable operating systems.

Support is already available as of Fedora 36. It's still being worked on. They may switch to using it as the default format for Fedora 38.

- F36: https://fedoraproject.org/wiki/Changes/OstreeNativeContainer
- F38: https://fedoraproject.org/wiki/Changes/OstreeNativeContainerStable
- CoreOS docs: 
   - https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md
   - https://coreos.github.io/rpm-ostree/container/
- Silverblue isn't officially supported yet, but it's coming soon. There's a thread at:
   - https://discussion.fedoraproject.org/t/feature-for-custom-rpm-ostree-container-native-builds/44480
   - Official should be coming soon. PR landed @ https://pagure.io/releng/pull-request/11120
   - Temporary Silverblue container (especially useful to do a `FROM` in a `Containerfile`): https://github.com/cgwalters/sync-fedora-ostree-containers
   - And third party builds are available at:
      - https://github.com/GuiltyDoggy/ostree-container
      - https://github.com/castrojo/ublue-image
- CoreOS has an official container for this and there are even third party ones now too.
   - Official: https://quay.io/repository/fedora/fedora-coreos?tab=tags
   - Examples: https://github.com/coreos/layering-examples
   - Third party: https://github.com/jlebon/pet/ 

Related, but a little different: https://github.com/containers/bootc (we'll likely keep our integration point at rpm-ostree)

garrett:
https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md#existing-work

> Specifically as of today, this functionality is exposed in:
> 
> - `rpm-ostree ex-container`
> - `rpm-ostree rebase --experimental $containerref`

HOWEVER, according to https://coreos.github.io/rpm-ostree/container/ it's even easier, perhaps?

> Rebasing a client system
> 
> Use this to switch to booting from a container image:
> 
> $ `rpm-ostree rebase ostree-unverified-registry:quay.io/fedora/fedora-coreos:stable
`

And:

[...]

PR #359, opened 2023-05-12
Fix crash on deployments without origin, initial OCI test

martinpitt:
First steps towards issue #300


martinpitt:
[Changed from local OCI dir to registry](https://github.com/cockpit-project/cockpit-ostree/compare/a548b3917c55122c877987e2ec9cf4176da076cd..b18f7257c91410222c8e7daf1ffc9659449a448e), which is much more realistic.

marusak (approved):
Thanks! Not sure what is the purpose of the `inputhash` but if it is needed...

martinpitt:
@marusak Heh, yes -- and also fairly undocumented. It took me almost an hour to get from "weird build crash" to this hack.. :grin:

PR #365, opened 2023-05-25
test: Initial fixes for ruff

martinpitt:
After we started using ruff in our projects, vim ALE now screams about too long lines everywhere, which is irritating. Configure it to use our standard 118 characters length, and fix the remaining overly long lines. Now `ruff check test/check-ostree` is clean.


allisonkarlitskaya (approved):
:+1:

PR #381, opened 2023-06-29
More ruff rules

jelly:
@martinpitt so I noticed we also have `make check-unit` but don't run that either in CI, we should probably do that?

I also noticed it npm install's stuff so that's something to fix.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 194
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ bots/
 pkg
 test/common/
 test/images/
+test/static-code
 *.mo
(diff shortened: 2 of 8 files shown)


martinpitt:
> but don't run that either in CI, we should probably do that?

We do -- `make check` depends on it, and you can see it at the bottom of the "initialization" section of our test logs:
```
npm run test

> test
> qunit src/test-*.js

TAP version 13
ok 1 parseData
ok 2 changeData
1..2
```

Incidentally, that's also the reason why the test count is completely messed up in these logs: 2 tests, 6 passed, 0 failed, 0 skipped, -4 to go  (it sees the unit test TAP plan first)

martinpitt (approved):
Cool, thank you!

PR #395, opened 2023-07-26
build.js: support flags

jelly:
Adds flags to build.js to use rsync, disable linting and use watch mode.


martinpitt (approved):
Thanks!

tomasmatus (dismissed):
This also needs README update

PR #397, opened 2023-07-27
test: add coverage scenario for ostree

jelly:


jelly:
Looks good! https://cockpit-logs.us-east-1.linodeobjects.com/pull-397-20230727-174152-0346255e-fedora-coreos-devel/Coverage/index.html

Makefile
```diff
@@ -100,11 +105,11 @@ print-version:
 dist: $(TARFILE)
 	@ls -1 $(TARFILE)
 
-# when building a distribution tarball, call bundler with a 'production' environment
+# when building a distribution tarball, call webpack with a 'production' environment by default
```
martinpitt, line 108: Please revert the "webpack" word, "bundler" is correct. This isn't using webpack.

martinpitt:
Nice! do we have any chance to exclude pkg/lib/ from the coverage report? *by design* we aren't going to use all shared components, and whenever we add a new one, the coverage would appear to go down.

martinpitt:
Once this lands, @jelly can you please file a bots PR about enabling the /devel scenario for all PRs?

jelly:
> Nice! do we have any chance to exclude pkg/lib/ from the coverage report? _by design_ we aren't going to use all shared components, and whenever we add a new one, the coverage would appear to go down.

We can, but iirc Marius liked including `pkg/lib` and we do so in machines/podman so I think we need to make a more general team decision :) (For stats it does make sense hmm to exclude it)

martinpitt (approved):
Yeah, it doesn't make sense to report coverage percentage for foreign code. We don't do this for anything in node_modules/ either, for example. Same thing for podman.

That said, this doesn't have to block this PR, thanks!

PR #407, opened 2023-08-23
packit: Add Fedora 39, drop Fedora 37

tomasmatus:


martinpitt (approved):
Thanks!

packit.yaml
```diff
@@ -41,21 +41,21 @@ jobs:
     metadata:
       dist_git_branches:
         - fedora-development
-        - fedora-37
         - fedora-38
+        - fedora-39
```
lachmanfrantisek, line 45: I wonder if any of [the existing aliases](https://packit.dev/docs/configuration/#aliases)  can't work for you so you don't need to update this twice a year... Sadly, I don't see any combination that would result in the same set of chroots. But maybe having `fedora-all` or `fedora-development`+`fedora-latest-stable` can work.

Maybe, we can create a new alias if really needed, but what is the rule here?  Is it `rawhide` plus two branched ones?

PR #417, opened 2023-10-02
Switch to dependabot for package.json updates

jelly:
Pin our devDependencies, as they will now be updated via dependabot in a more controlled fashion.

TIL: this project does not use a node_modules submodule


jelly:
```
TAP version 13
not ok 1 src/test-utils.js > Failed to load the test file with error:
Error [ERR_REQUIRE_ESM]: require() of ES Module /work/bots/make-checkout-workdir/src/test-utils.js from /work/bots/make-checkout-workdir/node_modules/qunit/src/cli/run.js not supported.
Instead change the require of test-utils.js in /work/bots/make-checkout-workdir/node_modules/qunit/src/cli/run.js to a dynamic import() which is available in all CommonJS modules.
    at run (/work/bots/make-checkout-workdir/node_modules/qunit/src/cli/run.js:55:4)
    at Object.<anonymous> (/work/bots/make-checkout-workdir/node_modules/qunit/bin/qunit.js:56:2)
  ---
  message: "should be able to load file"
  severity: failed
  actual: false
  expected: true
  stack:     at Object.<anonymous> (/work/bots/make-checkout-workdir/node_modules/qunit/src/cli/run.js:62:13)
  ...
1..1
# pass 0
# skip 0
# todo 0
# fail 1
make: *** [Makefile:167: check-unit] Error 1
```

martinpitt (approved):
Cheers!

PR #424, opened 2023-10-04
eslint: enable jsx-a11y plugin

jelly:
All our other projects enable this plugin to spot accessibility issues in JSX.


martinpitt (approved):
Thanks! It's past the build, seems to work. :+1:

PR #461, opened 2023-11-24
Drop running_method

jelly:
This no is no longer used since the switch to PatternFly. Uncovered by coverage!


martinpitt (approved):
:broom: nice!

PR #532, opened 2024-02-28
Drop old PF table actions override

jelly:
This was resolved in https://github.com/patternfly/patternfly-react/pull/5744

---

Manually verified, but hopefully pixel tests also cover it :)


martinpitt (approved):
Yay cleanup!

PR #541, opened 2024-03-11
Port Select to the non-deprecated variant

jelly:
The new Select is based on the Menu component as well which means we have to do a whole lot of things ourself like constructing the MenuToggle.


martinpitt:
Thanks @jelly ! At least this isn't half as bad as I feared. Do you think it makes sense to do this as a shared component again? FTR, that doesn't need to be a blocker -- if that's the whole thing, let's port c-ostree completely, and perhaps see how repetitive it'll be in podman etc.?

This needs to be rebased.

jelly:
Needs work, there are two other Select's I forgot to port over which are awfully the same.

src/repositoryModals.jsx
```diff
@@ -42,6 +43,34 @@ import * as remotes from './remotes';
 
 const _ = cockpit.gettext;
 
+const RemoteSelect = ({ remotes, selectedRemote, setSelectedRemote }) => {
+    const [isSelectOpen, setSelectOpen] = useState(false);
+
+    const toggle = toggleRef => (
+        <MenuToggle
+           ref={toggleRef}
+            onClick={() => setSelectOpen(!isSelectOpen)}
```
martinpitt, line 52: Indentation  -- why do we have all these fancy linters if they don't spot this?

martinpitt (approved):
Thanks for figuring this out! I suppose this should become a cockpit lib component (or even better, a standard PF one of course), that isSelectOpen/internal MenuToggle etc. logic is completely generic.

I leave it to you if you want to fix that indentation error before merging :wink:

src/repositoryModals.jsx
jelly: Yeah good question, I can look into this but that might change it everywhere I suppose.

jelly:
> Thanks for figuring this out! I suppose this should become a cockpit lib component (or even better, a standard PF one of course), that isSelectOpen/internal MenuToggle etc. logic is completely generic.

Yes, let's see how re-usable this becomes once we port more stuff :)

> I leave it to you if you want to fix that indentation error before merging 😉

I'll unfortunately leave it.

PR #548, opened 2024-03-14
Cleanup scss

jelly:
A bit of procastination as my other PR fails with pixel tests for RTL, so let's see if this one "flakes" or "fails" as well.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-541-6cfb7af5-20240314-095647-fedora-coreos/log.html


jelly:
Went green so the other PR needs more work.

martinpitt (approved):
Cheers!

PR #577, opened 2024-04-19
Initial test setup on centos-9-bootc

martinpitt:
This doesn't do much yet, but gives us something to build on.


martinpitt:
For the next steps I need to change our image, to actually ship a container image to upgrade to. This requires a bots change, but I'd like to land this now so that we don't break the existing functionality again on image refreshes.

allisonkarlitskaya (approved):
Cool stuff!

PR #579, opened 2024-04-22
test: Check rebasing bootc container image

martinpitt:
Since [1] the centos-9-bootc image contains an OCI image of its own
deployment. Build a "cockpit1" tag out of that which removes the `less`
package, so that we can more confidently assert which tag is deployed
and which one staged.

[1] https://github.com/cockpit-project/bots/pull/6243

---

 - [ ] That bots PR still needs to land


jelly (approved):
Cool stuff!

PR #638, opened 2024-06-20
build.js: write metafile in a normal build

jelly:
In the introduction of the metafile argument accidentally it was added to the watch mode instead of the normal build.

---

Don't hack in the airport lobby :grin:

### DIFF
--- a/build.js
+++ b/build.js
@@ -107,7 +107,10 @@ const context = await esbuild.context({
 });
 
 try {
-    await context.rebuild();
+    const result = await context.rebuild();
+    if (args.metafile) {
+        fs.writeFileSync(args.metafile, JSON.stringify(result.metafile));
+    }
 } catch (e) {
     if (!args.watch)
         process.exit(1);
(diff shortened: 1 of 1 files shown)


martinpitt (commented):
Ah, ok! But that means that metafiles will not be produced during watch mode? Don't we want both?

jelly:
> Ah, ok! But that means that metafiles will not be produced during watch mode? Don't we want both?

They don't, we could add it but it feels unneeded. Usually the builds are quick enough and you need to upload a file so getting new metadata is already a separate "step".

martinpitt (approved):
OK, fair enough. Do we need to copy this fix to all the other projects?

jelly:
> OK, fair enough. Do we need to copy this fix to all the other projects?

This was a mistaken introduced by failing to handle conflicts from `git am` all other projects are fine.

jelly:
Second bootc failure in a row, retrying one more time.

PR #647, opened 2024-07-03
eslint: configure import sorting rules

jelly:
This more or less enforces existing practice, but cleans up many places where we've deviated from that.  We sort into these groups:

- 'react', if present, always first
- external (`node_modules/`) imports
- cockpitlib (`pkg/lib/`) imports
- parent (`../`) imports
- sibling (`./`) imports

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,16 +1,16 @@
 #!/usr/bin/env node
 
 import fs from 'node:fs';
-import path from 'node:path';
 import os from 'node:os';
+import path from 'node:path';
 
 import copy from 'esbuild-plugin-copy';
 
-import { cleanPlugin } from './pkg/lib/esbuild-cleanup-plugin.js';
-import { cockpitCompressPlugin } from './pkg/lib/esbuild-compress-plugin.js';
 import { cockpitPoEsbuildPlugin } from './pkg/lib/cockpit-po-plugin.js';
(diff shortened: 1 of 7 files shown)


martinpitt:
Still some complaints about build.js

build.js
```diff
@@ -1,16 +1,16 @@
...
 
 import copy from 'esbuild-plugin-copy';
 
-import { cleanPlugin } from './pkg/lib/esbuild-cleanup-plugin.js';
-import { cockpitCompressPlugin } from './pkg/lib/esbuild-compress-plugin.js';
 import { cockpitPoEsbuildPlugin } from './pkg/lib/cockpit-po-plugin.js';
 import { cockpitRsyncEsbuildPlugin } from './pkg/lib/cockpit-rsync-plugin.js';
+import { cleanPlugin } from './pkg/lib/esbuild-cleanup-plugin.js';
 import { esbuildStylesPlugins } from './pkg/lib/esbuild-common.js';
+import { cockpitCompressPlugin } from './pkg/lib/esbuild-compress-plugin.js';
```
martinpitt, line 13: This is weird -- it was sorted before, and now it **un**sorted it?!

martinpitt: Oh, it sorted it by file name, not by imported names.. :man_shrugging:

martinpitt (approved):
Green → good, thanks!

PR #653, opened 2024-07-10
package.json: Explicitly depend on stylelint-config-recommended-scss

martinpitt:
The latest version 14.1.0 does not work with our other stylelint package versions. Pin down the version we've used so far, and let dependabot take care of grouped updates.

---

Fixes [this spontaneous breakage](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6598-f4894eed-20240709-224704-fedora-coreos-devel-cockpit-project-cockpit-ostree/log.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "sass": "1.77.6",
     "sizzle": "2.3.10",
     "stylelint": "16.6.1",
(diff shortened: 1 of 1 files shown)


tomasmatus (approved):
Thanks

PR #662, opened 2024-07-18
test: Robustify and clarify deployment selectors

martinpitt:
We really should avoid `tbody:nth-child(n)` selectors, as they are so
misleading: Reading that suggests that that this selects the nth
`<tbody>` tag, but that isn't so [1] -- that's what `:nth-of-type(n)` does.
That's why all of the numbers were apparently off by one, as they had to
account for the extra `<thead>` in the parent `<table>`.

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

----

This doesn't solve https://github.com/cockpit-project/bots/pull/6630 , but makes the investigation a lot more clear.

### DIFF
--- a/test/check-ostree
+++ b/test/check-ostree
@@ -36,7 +36,7 @@ KEY_ID = "95A8BA1754D0E95E2B3A98A7EE15015654780CBD"
 
 
 def switch_tab(b, index, tab):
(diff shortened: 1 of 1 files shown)


jelly (approved):
Thanks!

PR #665, opened 2024-07-18
Eliminate :contains

martinpitt:


src/ostree.jsx
```diff
@@ -629,7 +632,7 @@ const OStreeSource = ({ ostreeState, refreshRemotes, onChangeBranch, onChangeRem
     const Dialogs = useDialogs();
 
     const actions = [
-        <DropdownItem key="rebase"
+        <DropdownItem key="rebase" data-action="rebase"
```
jelly, line 635: Would be nice if the key leaked through to the DOM right ;-)

jelly (approved):
Assuming this goes green, nice!

I was looking at this briefly, and finding something for the dropdownitem seemed annoying, but `data-action` actually seems nice!

The tab solution is alright, not superb, but it doesn't seem like our table supports setting any id / description on the tab. Maybe something to consider but then again asserting that the tab position is 1,2,3 is already great! (And also a test of sorts)

jelly (commented):
Hmm what we just merged is flaky? That's strange!

martinpitt:
Ah no, much simpler -- `become_superuser()` requires `:contains()`, so it's dropping sizzle that broke it. I'll fix that tomorrow.

src/ostree.jsx
martinpitt: heh yea, but it's only a React detail..

martinpitt:
I dropped the "remove sizzle" commit for now. I'll make test/common free of :contains tomorrow, and can then drop sizzle for good.

jelly:
> I dropped the "remove sizzle" commit for now. I'll make test/common free of :contains tomorrow, and can then drop sizzle for good.

Aaaah, that explains a lot :) Was there a clue in the logs? :)

jelly (approved):
Hurray!

martinpitt:
The clue was to stare at

> self.wait_in_text("div[role=dialog]:contains('Switch to administrative access')", f"Password for {user}:")

for a few seconds and realize the `:contains` :grin:

PR #679, opened 2024-08-13
test: Move from CDP to BiDi

martinpitt:
See https://github.com/cockpit-project/cockpit/pull/20832

Explicitly add a "glob" devDependency to follow suit with https://github.com/cockpit-project/cockpit/commit/680decc155a It was previously used implicitly through a transient dependency of something else, but our esbuild po-plugin uses it explicitly.

 - [x] https://github.com/cockpit-project/cockpit/pull/20883

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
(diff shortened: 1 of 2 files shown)


martinpitt:
[testPermission](https://cockpit-logs.us-east-1.linodeobjects.com/pull-679-17a4ea63-20240813-103128-fedora-coreos/log.html) should be fixed in webdriver_bidi.py. I have the fix, but I'm collecting some other bidi follow-ups.

The pixel test is plain buggy, need to investigate.

tomasmatus (approved):

PR #688, opened 2024-08-27
Use rpm-ostree's RegisterClient() method for stopping suicide

martinpitt:
This is a better workaroud for
https://issues.redhat.com/browse/RHEL-49603
than the hack commit e20543c79575e042b284. Revert the latter.

---

I tested this locally with `ExitIdleTimeout=10`. I just don't want to do that in the VM setup, and rather keep the default config.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -176,6 +176,11 @@ class RPMOSTreeDBusClient {
             this.sysroot.addEventListener("changed", this.on_sysroot_changed);
             this.sysroot.wait(() => {
                 if (this.client) {
+                    /* HACK: by default, rpm-ostreed's IdleTimeout is way too aggressive and broken
+                    * Tell it to not do that */
+                    this.sysroot.RegisterClient({ id: cockpit.variant("s", "cockpit-ostree") })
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -176,6 +176,11 @@ class RPMOSTreeDBusClient {
             this.sysroot.addEventListener("changed", this.on_sysroot_changed);
             this.sysroot.wait(() => {
                 if (this.client) {
+                    /* HACK: by default, rpm-ostreed's IdleTimeout is way too aggressive and broken
+                    * Tell it to not do that */
+                    this.sysroot.RegisterClient({ id: cockpit.variant("s", "cockpit-ostree") })
+                            .catch(ex => console.error("Failed to register client:", JSON.stringify(ex)));
```
cockpituous, line 182: This added line is not executed by any test.

mvollmer (approved):
Ok, if that is how this is supposed to work...

martinpitt:
@mvollmer That's not what I consider a good behaviour, but it's what Colin Walters told me in https://issues.redhat.com/browse/RHEL-49603 :grin:

PR #702, opened 2024-09-19
spec: Do not build on ix86

martinpitt:
cockpit-ostree is one of the dependents of rpm-ostree. We are removing i686 buids on this pkg from fedora version 41 and above.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1716#issuecomment-2358745869

---

Originally submitted here: https://src.fedoraproject.org/rpms/cockpit-ostree/pull-request/52


martinpitt:
LGTM, but need someone else to formally approve.

jelly (approved):

travier:
I've just realized that this is a noarch package, so there is no need to mark it as not for a specific architecture (at least right now, it could grow arch specific code).

PR #787, opened 2025-02-07
package.json: drop bootstrap-sass dependency

jelly:
The ostree plugin doesn't use bootstrap anymore since the React/PF4 rewrite in 2020.


martinpitt (approved):
Odd, how did the NPM mirror work in this PR, but not the others..

PR #809, opened 2025-03-24
Migration to Patternfly v6

tomasmatus:


src/deploymentModals.jsx
```diff
@@ -73,7 +73,7 @@ export const CleanUpModal = ({ os }) => {
     const actions = [
         <Button key="cleanup-accept"
             variant="primary"
-            isDisabled={buttonLoading || (!deleteTemporaryFiles && !deleteRPMmetadata && !deletePendingDeployments && !deleteRollbackDeployments)}
+            isAriaDisabled={buttonLoading || (!deleteTemporaryFiles && !deleteRPMmetadata && !deletePendingDeployments && !deleteRollbackDeployments)}
```
cockpituous, line 76: This added line is not executed by any test.

tomasmatus:
Pixeltests look quite decent

src/deploymentModals.jsx
```diff
@@ -22,8 +22,8 @@ import React, { useState } from 'react';
 import { Alert } from "@patternfly/react-core/dist/esm/components/Alert";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Checkbox } from '@patternfly/react-core/dist/esm/components/Checkbox';
-import { Modal } from "@patternfly/react-core/dist/esm/components/Modal";
-import { Text } from "@patternfly/react-core/dist/esm/components/Text";
+import { Content } from "@patternfly/react-core/dist/esm/components/Content";
+import { Modal } from '@patternfly/react-core/dist/esm/deprecated/components/Modal';
```
martinpitt, line 26: :eyes: modals are deprecated? Or there's a new one with different API?

martinpitt:
@tomasmatus I agree, thanks for porting! Please squash, push the pixels, and :rocket:

src/deploymentModals.jsx
tomasmatus: Yes, there is a new modal in pf6 that supports fewer props and there are new sub-components `ModalHeader`, `ModalBody`, `ModalFooter`

src/deploymentModals.jsx
cockpituous, line 76: This added line is not executed by any test.

src/repositoryModals.jsx
```diff
@@ -393,14 +394,20 @@ export const RebaseRepositoryModal = ({ origin, availableRemotes, currentOriginB
             />
         )
         : (
-            <Text>{availableRemotes[0]}</Text>
+            <Content component="p">{availableRemotes[0]}</Content>
```
cockpituous, line 397: This added line is not executed by any test.

src/deploymentModals.jsx
martinpitt: This needs some explanation. Why is `isDisabled` now out of fashion? We still use it in cockpit.git. AFAIUI, the aria properties should only be used if the regular properties somehow don't exist -- aria is a hack/workaround, proper semantic HTML is better.

martinpitt (commented):
Thanks! I'd like to know about the aria stuff, as that feels dubious to me. Otherwise I'll leave the final review to @Venefilyn for sign-off.

src/deploymentModals.jsx
Venefilyn: @martinpitt it's for accessibility reasons. To quote [Patternfly website](https://www.patternfly.org/components/button/accessibility)

> If a button should be disabled, but remain focusable, it has the `aria-disabled="true"` attribute instead of the standard `disabled` attribute.

martinpitt: This feels upside down to me, but I'll bow to the expert :grin:  Thanks!

martinpitt (approved):

src/deploymentModals.jsx
Venefilyn: @martinpitt You can try it on their website too so you can see it in action with keyboard navigation

Try to tab to the [disabled buttons here](https://www.patternfly.org/components/button#disabled-buttons) and it won't work, but [aria disabled buttons here](https://www.patternfly.org/components/button#aria-disabled-examples) you can focus

martinpitt: Right, but is that desirable? We e.g. don't make the disabled link buttons tab-able on the Overview page either. Maybe that's desired, maybe not. I don't know, so I'll just let you decide. At least I now now what the difference is, thanks!

PR #810, opened 2025-03-25
test: obtain bootc booted version from json

jelly:
bootc changed the version output in `bootc status` to now quote the version number. This breaks our test which expects an unquoted string.


martinpitt (approved):
jq FTW!

PR #822, opened 2025-04-09
update pixel referece

tomasmatus:
fixes: https://github.com/cockpit-project/bots/pull/7636

Font slightly changed as we now include correct PF6 fonts.


tomasmatus:
@jelly I think I need help:

pixeltest passes when I trigger the tests with `./tests-trigger --repo cockpit-project/cockpit-ostree 822 fedora-coreos@bots#7636`

but fails when I triggered it with `bots/tests-trigger --repo cockpit-project/cockpit-ostree --bots-pr 7636 822 image:fedora-coreos`
Here it fails because somehow it still looks like it is using the old image without new fonts?

jelly:
> @jelly I think I need help:
> 
> pixeltest passes when I trigger the tests with `./tests-trigger --repo cockpit-project/cockpit-ostree 822 fedora-coreos@bots#7636`
> 
> but fails when I triggered it with `bots/tests-trigger --repo cockpit-project/cockpit-ostree --bots-pr 7636 822 image:fedora-coreos` Here it fails because somehow it still looks like it is using the old image without new fonts?

Maybe we should first merge https://github.com/cockpit-project/cockpit-ostree/pull/823 and then retry this one?

PR #842, opened 2025-05-12
test: update pixel tests for cockpit-ws 337

jelly:
cockpit-ws now serves the new font which comes with PatternFly 6.

### DIFF
--- a/src/ostree.scss
+++ b/src/ostree.scss
@@ -65,3 +65,11 @@
 #cleanup-deployment-modal .pf-v6-c-check:not(:last-of-type) {
     margin-block-end: var(--pf-t--global--spacer--xs);
 }
+
+// PF6 workaround, when table is in the small screen layout PF sets too big font-weight
+// https://github.com/patternfly/patternfly/issues/7526
+@media screen and (max-width: $pf-v6-global--breakpoint--lg) {
+    #deployments .pf-v6-c-table :where(.pf-v6-c-table__th, .pf-v6-c-table__td)[data-label]::before {
+        font-weight: var(--pf-t--global--font--weight--body--bold);
+    }
+}


tomasmatus:
the `OstreeRestartCase-testOstree-deployment-mobile-pixels.png` seems to be too bold in the top part, I'll have a look
image

Venefilyn:
> the `OstreeRestartCase-testOstree-deployment-mobile-pixels.png` seems to be too bold in the top part, I'll have a look image

Yeah that seems to be the only thing I can spot too. I did notice some headings losing their boldness but it looks fine IMO and is better anyway

- Version/Status/Time/Branch likely has an override that is in the 700-800s and that is too much - could probably be removed and it'll look better
- `cockpit-base.1` has a bold that seems fine but we might wanna dial it down? Or drop completely

tomasmatus:
> Version/Status/Time/Branch likely has an override that is in the 700-800s and that is too much - could probably be removed and it'll look better

seems like that comes from patternfly directly as in devtools I can see the following:
image

I can override it but probably worth reporting to PF

> cockpit-base.1 has a bold that seems fine but we might wanna dial it down? Or drop completely

seems fine to me, it has weight 500 which is the lowest of the bold fonts

tomasmatus:
This should fix the font-weight problem
image

src/ostree.scss
```diff
@@ -65,3 +65,10 @@
 #cleanup-deployment-modal .pf-v6-c-check:not(:last-of-type) {
     margin-block-end: var(--pf-t--global--spacer--xs);
 }
+
+// PF6 workaround, when table is in the small screen layout PF sets too big font-weight
+@media screen and (max-width: $pf-v6-global--breakpoint--lg) {
+    #deployments .pf-v6-c-table :where(.pf-v6-c-table__th, .pf-v6-c-table__td)[data-label]::before {
+        font-weight: var(--pf-t--global--font--weight--body--bold);
+    }
+}
```
Venefilyn, line 74: This seems like it is something we should have in Cockpit PF workarounds file

tomasmatus (approved):
Assuming green 👍🏻

src/ostree.scss
tomasmatus: Let's do this separately in order to not block this image refresh

jelly:
Went green(!), but now we need to land the image or this PR first.

PR #880, opened 2025-07-01
test: Adjust pixel ref for image update

martinpitt:
The fcos update in https://github.com/cockpit-project/bots/pull/7938 persistently and reliably changes the spacing on the deployment page. Why it does that is completely beyond me, but let's just accept the new reality.

----

See [failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7938-f00ed3c0-20250701-122353-fedora-coreos-cockpit-project-cockpit-ostree/log.html). I've ran it 5 times and it always changes the same way. Who knows, some invisible UUIDs in the DOM or whatnot :man_shrugging:


jelly (approved):

Venefilyn:
@martinpitt `fedora-coreos` is required to merge though, and we don't have override here. So how are we supposed to merge this one first?

PR #1090, opened 2026-06-04
test: Remove assumptions about pre-existing remotes

mvollmer:
By removing all but our "local" one.

Recent fedora-coreos images do not have the traditional "fedora" and "fedora-compose" remotes anymore, for unknown reasons.


jelly (approved):
Sure!

PR #1137, opened 2026-08-31
package.json: Drop esbuild-plugin-replace

jelly:
No longer needed since a year ago, esbuild-common.js no longer depends on it. See:

https://github.com/cockpit-project/cockpit/commit/5795ca9e09249192e4329517511795f49c9ae905


mvollmer (approved):
Thanks!

